PR #3, opened 2023-12-18
Fix random sim fails, add assert messages

andreasWallner:
StreamDriver does not drive the ready signal before the first clock edge, which can trigger the assertion in the LsuCachelessPlugin. Drive at simulation start to prevent that.

Add assert message to make it easier to debug regression fails caused by them.


Dolu1990:
Ahhhh lolol
Thanks :D

PR #4, opened 2023-12-24
Add iterative shifter

andreasWallner:
Reproduce issue with
`Test/runMain vexiiriscv.tester.TestBench --load-elf ext/NaxSoftware/baremetal/coremark/build/rv32ima/coremark.elf --with-konata --with-wave --with-spike-log --with-rvls-log --decoders=2 --lanes=2 --allow-bypass-from=0 --with-btb --with-late-alu --with-mul --with-div --with-iterative-shift`

### DIFF
--- a/src/main/scala/vexiiriscv/execute/IterativeShiftPlugin.scala
+++ b/src/main/scala/vexiiriscv/execute/IterativeShiftPlugin.scala
@@ -82,15 +82,22 @@ class IterativeShifterPlugin(val layer: LaneLayer,
 
       val amplitudeWidth = if(Riscv.XLEN.get == 64) 6 else 5
       val shamt = srcp.SRC2.resize(amplitudeWidth).asUInt
-      if(Riscv.XLEN.get==64) when(IS_W) {
-        amplitude(5) := False
-      }
       val rs1 = el(IntRegFile, RS1).asBits
 
       val busy = RegInit(False)
       val flipped = Reg(Bool())
-      val amplitude = Reg(cloneOf(shamt))
-      val shiftReg = Reg(cloneOf(srcp.SRC1.asBits))
+      val amplitude = Reg(UInt(amplitudeWidth bits))
+      val shiftReg = Reg(Bits(Riscv.XLEN bits))
+
+      val dataIn = CombInit(rs1)
+      if (Riscv.XLEN.get == 64) {
+        when(IS_W) {
+          shamt(5) := False
+        }
+        when(IS_W_RIGHT) {
+          dataIn(63 downto 32) := (default -> (ARITHMETIC & rs1(31)))
+        }
+      }
 
       val done = if(lateResult) {
         amplitude === 0 & !flipped & busy
@@ -106,12 +113,6 @@ class IterativeShifterPlugin(val layer: LaneLayer,
         // done comes one cycle "early", in the cycle we do the last action (shift / flip)
         shamt === 0 | leftShiftDone | rightShiftDone
       }
-
-      val signExtended =
-        if(Riscv.XLEN.get == 64)
-          IS_W_RIGHT ? (B(32 bit, default -> (ARITHMETIC & rs1(31))) ## rs1(31 downto 0)) | rs1
-        else
-          CombInit(rs1)
 
       val muxInputs = mutable.ArrayBuffer[(Bool, Bits, UInt)]()
       rightShifts.sorted.foreach(n => {
@@ -141,7 +142,7 @@ class IterativeShifterPlugin(val layer: LaneLayer,
           muxInputs.append((IS_W & doFlip, False #* 32 ## shiftReg(31 downto 0).reversed, amplitude))
         }
       }
-      muxInputs.append((selected & !busy, signExtended, shamt))
+      muxInputs.append((selected & !busy, dataIn, shamt))
 
       val selector = OHMasking.last(Cat(muxInputs.map(_._1)))
       val muxed = MuxOH(selector, muxInputs.map(_._2))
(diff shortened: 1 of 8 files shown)


src/main/scala/vexiiriscv/execute/IterativeShiftPlugin.scala
```diff
@@ -0,0 +1,109 @@
...
+          IS_W_RIGHT ? srcp.SRC1(31).asSInt.resize(32).resize(64).asBits | srcp.SRC1.asBits
+        else
+          srcp.SRC1.asBits
+
+      val shiftReg = Reg(cloneOf(srcp.SRC1.asBits))
+      val shifted = cloneOf(srcp.SRC1.asBits)
+      val shiftInput = busy ? shiftReg | signExtended
+      val shiftResult = zeroShift ? signExtended | shifted
+
+      shifted := (LEFT & True).mux(
```
Dolu1990, line 83: I'm asking myself if maybe we could do a "smarter" design with the following hardware shifts : 
- \>\> 1
- \>\> 8
- reverse all bits
- shift by zero case being emulated by doing 2 "reverse all bits"

That way we would have as many mux than now, but much high performance for big shifts, as we could compose shifts using the \>\> 8 and \ >\> 1

In a more generic manner, the number of different >> X possible could be a parameter List[Int]

So, the design would be kinda much closer to the actual BarrelShifterPlugin,  but over multiple cycles.

src/main/scala/vexiiriscv/execute/IterativeShiftPlugin.scala
```diff
@@ -0,0 +1,109 @@
...
+      val selected = isValid && SEL
+
+      val amplitudeWidth = if(Riscv.XLEN.get == 64) 6 else 5
+      val shamt = srcp.SRC2.resize(amplitudeWidth bit).asUInt
+      val zeroShift = shamt === 0
+
+      val busy = RegInit(False)
+      val amplitudeReg = Reg(cloneOf(shamt))
+      val amplitude = busy ? amplitudeReg | shamt
+      val done = amplitude(4 downto 1) === 0
```
Dolu1990, line 70: only for 32 bits it works ^^

src/main/scala/vexiiriscv/tester/TestBench.scala
```diff
@@ -456,7 +459,7 @@ object TestBench extends App{
     genConfig.includeSimulation
 
     val simConfig = SpinalSimConfig()
-    simConfig.withFstWave
+    simConfig.withVcdWave
```
Dolu1990, line 462: VCD wave are very dangerous as a default, as they can easily blow up a hard drive very first
```suggestion
    simConfig.withFstWave
```

src/main/scala/vexiiriscv/execute/IterativeShiftPlugin.scala
```diff
@@ -0,0 +1,201 @@
...
+      val muxInputs = mutable.ArrayBuffer[(Bool, Bits, UInt)]()
+      rightShifts.sorted.foreach(n => {
+        val doIt = if(n > 1) amplitude >= n else True
+        val input = (((ARITHMETIC && srcp.SRC1.msb) #* n) ## shiftReg) >> n
+        val ampl = amplitude - n
+        muxInputs.append((doIt, input, ampl))
+      })
+
+      leftShifts.sorted.foreach(n => {
+        // mask away bits shifted into upper 32 bit in case of SLLW instruction
```
Dolu1990, line 127: Is it necessary ? As anyway, the `ifp.signExtend(wb, layer(op), 32)` will erase all upper bits using the bit 31 as bit sign.

src/main/scala/vexiiriscv/execute/IterativeShiftPlugin.scala
```diff
@@ -0,0 +1,201 @@
...
+
+      SHIFT_RESULT := (if (lateResult) shiftReg else muxed)
+    }
+
+    val format = new el.Execute(formatAt) {
+      wb.valid := SEL
+      wb.payload := SHIFT_RESULT
+      if(Riscv.XLEN.get == 64) {
+        when(ARITHMETIC & IS_W) {
+          wb.payload(32, 32 bit).setAllTo(SHIFT_RESULT(31))
```
Dolu1990, line 196: This seems redondant with the `ifp.signExtend(wb, layer(op), 32)` 

ifp => IntFormatPlugin => a plugin which provide a integer writeback path with optional static sign/zero extends

andreasWallner: right, done^

src/main/scala/vexiiriscv/tester/TestBench.scala
Dolu1990: I guess overall, we would need to add all of those through argument / conifiguration / envrionnement variables ^^

I also have a lot the habit to modify / comment code in order to fit my specific needs, but overall, that doesn't scale up well XD

andreasWallner: that's why I often use this one: https://github.com/andreasWallner/spinalStuff/blob/90604fce38fab63d7a366e91a158b6e6f03e56e7/src/main/scala/andreasWallner/sim/package.scala#L268 ;-)
(though I'm not that happy with the name anymore)

src/main/scala/vexiiriscv/execute/IterativeShiftPlugin.scala
andreasWallner: fixed

Dolu1990:
Thanks :)

PR #11, opened 2024-04-02
Bitmanip

andreasWallner:
This should be an initial version thats mergeable, but not fully done yet.

Current TODO list that I'd open as an issue:
- Push changes to NaxSoftware so that we can have proper benchmarks - I'd not go for full combinatorics here, that would explode the number of images that we need, also needs a few changes in the regression (also the reason why I'm not yet adding the arch flags)
- optimize the current ZbbCountPlugin, the comb path should be horrific ATM
- possibly make the current situation with `ZbPlugin.make` cleaner, though not sure how... Zbb should not be a single plugin IMO


Dolu1990:
Soooo many plugins <3

I sent you rights to the NaxSoftware repo.
Feel free to merge :)

Thanks :D

Dolu1990:
>  I'd not go for full combinatorics here

Right ^^

So, fondamentaly, as long as there is no coupling between sub parts of the extension, i would say no need to handle them separatly in the binaries.

andreasWallner:
In the number of plugins: I kept them separate where there is no real reuse of hardware, so that a plugin is useful on its own and there is no big overlap...

On the combinatorics: to some degree there is overlap in the benchmarks...

Dolu1990:
> On the combinatorics: to some degree there is overlap in the benchmarks...

Right XD
I would say, let's keep full B extentions for now

PR #47, opened 2025-03-21
Scala 2.13.12 compatibility

Nik-Sch:
I fixed everything to support scala 2.13.12 and still be compatible with 2.12.18. This mainly includes the following changes:
- `Unit` -> `()`
- The default `Seq` is now `scala.collection.immutable.Seq`, therefore, I changed the definitions to `scala.collection.Seq` where applicable.
- In some circumstances the dotted notation was required (mainly for the `unbounded()` for `Opt[_]` and some `simPublic()`)
- Sometimes `toSeq` needed to be called for sequences to be explicitly immutable (e.g. in `ExecuteLanePlugin.scala:103` for the correct `retains` overload being used)

I am not familiar enough with `sbt` and the github workflows to add a unit test for scala 2.13. Could you give me some pointers how this could be added?


Tectu:
Looks like somebody was faster than me :)

Dolu1990:
Nice :D

So, for scala 2.13 regression. Maybe one example is SpinalHDL itself, but probably a bit complicated.
- It specifies all  the scala version here https://github.com/SpinalHDL/SpinalHDL/blob/dev/project/version.conf#L1
- then https://github.com/SpinalHDL/SpinalHDL/blob/dev/project/Version.scala#L11
- then https://github.com/SpinalHDL/SpinalHDL/blob/dev/build.sbt#L11
- then github action : https://github.com/SpinalHDL/SpinalHDL/blob/dev/.github/workflows/run-tests-2_13.yml

But i'm i didn't puted it in place myself, so i may miss something

Dolu1990:
That was an issue with Scala 2.13 breaking user space API <3<3<3
Fixed now. let's see how it goes.

PR #70, opened 2025-08-18
Add Sstc support

inochisa:
First, add conditional CSR filter to allow access CSR with extra conditions.

Then add Sstc support for RISC-V based on the ISA Manual.

Test with litex SoC


Dolu1990:
Nice thanks :D

PR #74, opened 2025-08-25
Trigger interrupt based on AIA requirement

inochisa:
AIA defined interrupt order for machine and supervisor interrupt. Implement this can give a predictable trigger order for interrupt.

This should have no side effect. Refer chapter 5 of the [aia spec](https://github.com/riscv/riscv-aia) for the order it describe.

PR #77, opened 2025-08-31
Introduce basic AIA support

inochisa:
The ISA part of the RISC-V AIA includes three things:
1. Indirect CSR support (`Smcsrind`, `Sscsrind`).
2. `xTOPI` CSR (`MTOPI`, `STOPI`)
3. Interrupt filter

This PR includes the first two things. The interrupt filter requires some hard change for `PrivilegedPlugin` and I am still trying to figure it out. However, this PR is enough for almost all the case, as the interrupt filter is not used at least for now (For OpenSBI and Linux). only `xTOPI` CSR is required.

Also, As the  `xTOPI` CSR just exports the information from PR #74, no new option is introduced to mask this CSR.

### DIFF
--- a/src/main/scala/vexiiriscv/Param.scala
+++ b/src/main/scala/vexiiriscv/Param.scala
@@ -136,6 +136,7 @@ class ParamSimple() {
   var withRvd = false
   var withRvZb = false
   var withWhiteboxerOutputs = false
+  var withIndirectCsr = false
   var privParam = PrivilegedParam.base
   var lsuForkAt = 0
   var lsuPmaAt = 0
@@ -500,6 +501,7 @@ class ParamSimple() {
     if (withRvd) isa += "d"
     if (withRvc) isa += "c"
     if (withRvZb) isa += "ZbaZbbZbcZbs"
+    if (withIndirectCsr) isa += "Smcsrind" + privParam.withSupervisor.mux("Sscsrind", "")
     if (privParam.withSupervisor) isa += "s"
     if (privParam.withUser) isa += "u"
     val r = new ArrayBuffer[String]()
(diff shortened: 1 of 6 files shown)


src/main/scala/vexiiriscv/Param.scala
```diff
@@ -987,6 +987,7 @@ class ParamSimple() {
     plugins += new CsrRamPlugin()
     if(withPerformanceCounters) plugins += new PerformanceCounterPlugin(additionalCounterCount = additionalPerformanceCounters)
     plugins += new CsrAccessPlugin(early0, writeBackKey =  if(lanes == 1) "lane0" else "lane1")
+    plugins += new IndirectCsrPlugin(privParam.withSupervisor)
```
Dolu1990, line 990: This should be a disabled by default extention.
The reason is to keep VexiiRiscv small on FPGA, most optional feature are disabled. (it addup quikly ^^)

src/main/scala/vexiiriscv/misc/IndirectCsrPlugin.scala
```diff
@@ -0,0 +1,57 @@
...
+    val buildBefore = retains(cap.csrLock)
+
+    awaitBuild()
+
+    val harts = for (hartId <- 0 until HART_COUNT) yield new Area {
+      val api = cap.hart(hartId)
+
+      val m = new Area {
+        val iselect = RegInit(U(0, XLEN bits))
+        api.readWrite(iselect, CSR.MISELECT)
```
Dolu1990, line 30: To ilustrate why plugins += new IndirectCsrPlugin(privParam.withSupervisor) should be optional (and disable by default)

```scala
        val iselect = RegInit(U(0, XLEN bits))
        api.readWrite(iselect, CSR.MISELECT)
``` 

is expensive

inochisa:
> Hi ^^ Thanks :D
> 
> > only xTOPI CSR is required.
> 
> required in which context ?

Required in the context when the software wants to use `Smaia` or `Ssaia` extension. The kernel and OpenSBI always uses it as an fast way to access interrupt information if this extension is supported.

inochisa:
Updated

inochisa:
@Dolu1990

Dolu1990:
Hi, 

Sorry i missed your updates.
I'm generaly swamped in work those weeks, spam me if i'm unreactive ^^

> I have a small question for make it optional: if another feature require this plugin, should it be selected in silence, or an error will be thrown?

You can select it in silence, that is fine 

Cheers
Charles

inochisa:
> Hi,
> 
> Sorry i missed your updates. I'm generaly swamped in work those weeks, spam me if i'm unreactive ^^
> 

There is no need to rush. Two weeks is not too long for me.

> > I have a small question for make it optional: if another feature require this plugin, should it be selected in silence, or an error will be thrown?
> 
> You can select it in silence, that is fine
> 

Thanks, it is helpful for me.

PR #95, opened 2025-11-12
Add AIA interrupt filter function

inochisa:
Add virtual interrupt and interrupt filter support, which is the last part for the full function AIA support.

Since I found no one uses this feature, I write a simple test by myself:
https://github.com/project-inochi/NaxSoftware/tree/test/baremetal/interrupt_filter


inochisa:
Hi, @Dolu1990 , could you take a look at this PR?

PR #100, opened 2025-12-15
Use separate FSM state for MMU doneLogic

inochisa:
Using Stream instead of Flow for the MMU PTW load logic, which has two benifits:
1. Allow the MMU to hold the load respond value with desired time, instead of only one cycle.
2. Reduce the logic cell usage.

With this PR, I found about 200 LUTs is saved for each core, and no obvious performance impact is observed:

```
coremark Iterations/Sec
With this PR       : 222.485909
without this PR    : 222.403440
```


Dolu1990:
Thanks ^^

PR #120, opened 2026-02-21
Add initital H extension support (Without Full AIA support)

inochisa:
Add basic support for H extension.

This PR includes the following part:
1. introduce a new method to remap the VS-stage S-mode CSR access to VS-stage, also add necessary api for HS-mode permission check
2. Add basic hypervisor CSR support.
3. Adjust TrapPlugin to support VS-stage interrupt.
4. Adapt the sstc/smcsrind/sscsrind to hypervisor extension
5. G-stage MMU support

This PR does not include:
1. ~~G-stage MMU support.~~
2. Full AIA extension support for H extension.

Edit:
20260316: Add G-stage MMU support


inochisa:
The impact of this PR  (1 core with PLIC) 

Compile options: `--build --update-repo no --cpu-type vexiiriscv --cpu-variant debian --with-coherent-dma --no-netlist-cache --with-ethernet --libc-mode full` The H extension is enabled by `--with-isa h`

## LUTs

This are compiled on XCKU5P

| Ref | LUTs |
|:--------:|:----------:|
| HEAD          | 17533 |
| H enabled  | 19687 |
| H disabled | 18075 |

## Performance

Coremark:

| Ref | Iterations/Sec |
|:--------:|:----------:|
| HEAD          | 208.130984 |
| H disabled | 207.569363 |
| H enabled (HS) | 186.811134 |
| H enabled (VS) | 136.425648 |

It seems like about 10% performance from this PR **with H extension enabled** , and this mainly comes from a new pipeline stage in both LSU and Fetch.

Edit: The performance impact on S mode is removed by only inserting the stage-2 translation with extension enabled.
Edit: Add VS mode coremark score.

inochisa:
Hi, @Dolu1990 

Currently, I think this PR is worth to review now. At least It can boot Linux and run a simple bare KVM program (I still try to setup an rootfs so the Qemu can be used to boot a linux in KVM mode). It may still have some small bugs but there should be no big change anymore.

Dolu1990:
Hi,

gg ^^

For the performance, my guess is that in a ASIC it would be totaly fine to chain both mmu translation in a single cycle. So FPGA degradation only.

Two things i can see : 
- src/main/scala/vexiiriscv/Param.scala has the pipeline stages  + 1 hard coded
- Would need to integrate regressions tests for H support in https://github.com/SpinalHDL/VexiiRiscv/blob/dev/src/test/scala/vexiiriscv/tester/RegressionSingle.scala
- Would then need to specify how to generate random H generation parmeters in  https://github.com/SpinalHDL/VexiiRiscv/blob/dev/src/test/scala/vexiiriscv/tester/Regression.scala 

That way we can run the regressions tests for configs which have H support ^^
Idealy, this should include booting linux in linux in a simulation. Doing it in simulation would allow to catch any spec missmatch against spike which is exercised by the whole process.
Also, this would allow to do changes without too much fear of breaking things without noticing.

Dolu1990:
Note, coremark is quite sensitive to LSU latency, as it does quite a bit of linked list pointer chassing.

inochisa:
> Hi,
> 
> gg ^^
> 
> For the performance, my guess is that in a ASIC it would be totaly fine to chain both mmu translation in a single cycle. So FPGA degradation only.
> 
> Two things i can see :
> 
>     * src/main/scala/vexiiriscv/Param.scala has the pipeline stages  + 1 hard coded
> 

Yes, it is hard coded now. However, for non-H mode build, the new stage could be emitted. As the `stpk` in Fetch/LSU is damaged and not accessed at all in this scene. However, it is ugly for using something like `if (withRvh) X + 1 else X` for the parameters. So I guess we may need to define some template for it.

>     * Would need to integrate regressions tests for H support in https://github.com/SpinalHDL/VexiiRiscv/blob/dev/src/test/scala/vexiiriscv/tester/RegressionSingle.scala
> 
>     * Would then need to specify how to generate random H generation parmeters in  https://github.com/SpinalHDL/VexiiRiscv/blob/dev/src/test/scala/vexiiriscv/tester/Regression.scala
> 
> 
> That way we can run the regressions tests for configs which have H support ^^ 

Yes, that's true. It is fine for me to add it.

> Idealy, this should include booting linux in linux in a simulation.

Linux could be a problem, not for a synthesized core, but for the simulation. In fact, I managed to boot an mainline OpenSBI with the simulation. And it gave me a 1.1GB trace log and a 5.1 GB fst wave. XD

And at least for now, my team are struggled to boot a linux with rootfs. Since my board has no SD card and I have to boot the kernel with ethernet.

> Doing it in simulation would allow to catch any spec missmatch against spike which is exercised by the whole process. Also, this would allow to do changes without too much fear of breaking things without noticing.

My suggestion is adding a batch of unit tests like [defermelowie/riscv-hext-asm-tests](https://github.com/defermelowie/riscv-hext-asm-tests) (And I just use this to test the implementation). So we can test all the spec requirements with both log and wave. (By this way, I think the spike for Vexiiriscv is too old to support H extension. XD)

With lots of unit test are required, I think a small framework and some common code are needed to invoke the unit test automatically. Now I used `TestBenchServer` for this, but it is hard to capture the test state. Anyway, this is more like the thing in the future.

inochisa:
And a good news, The kvm unit test has passed.

```
BUILD_HEAD=86e53277
timeout -k 1s --foreground 90s /usr/bin/lkvm run --kernel /tmp/tmp.nxeSLOa1c6 --cpus 1 --nodefaults --network mode=none --loglevel=warning # --initrd /tmp/tmp.YMA4zI7XFd

##########################################################################
#    kvm-unit-tests
##########################################################################

PASS: sbi: base: Bad FID: SBI_ERR_NOT_SUPPORTED
SKIP: sbi: base: spec_version: missing SBI_SPEC_VERSION environment variable
SKIP: sbi: base: impl_id: missing SBI_IMPL_ID environment variable
SKIP: sbi: base: impl_version: missing SBI_IMPL_VERSION environment variable
PASS: sbi: base: probe_ext: check sbi.error and sbi.value
PASS: sbi: base: probe_ext: unavailable: check sbi.error and sbi.value
SKIP: sbi: base: mvendorid: missing MVENDORID environment variable
SKIP: sbi: base: marchid: missing MARCHID environment variable
SKIP: sbi: base: mimpid: missing MIMPID environment variable
PASS: sbi: time: Bad FID: SBI_ERR_NOT_SUPPORTED
PASS: sbi: time: set_timer: set timer
PASS: sbi: time: set_timer: timer interrupt received
PASS: sbi: time: set_timer: pending timer interrupt bit set in irq handler
PASS: sbi: time: set_timer: pending timer interrupt bit cleared by setting timer to -1
PASS: sbi: time: set_timer: timer delay honored
PASS: sbi: time: set_timer: timer interrupt received exactly once
PASS: sbi: time: set_timer: set timer for mask irq test
PASS: sbi: time: set_timer: timer interrupt received for mask irq test
PASS: sbi: time: set_timer: pending timer interrupt bit set in irq handler for mask irq test
PASS: sbi: time: set_timer: timer delay honored for mask irq test
PASS: sbi: time: set_timer: timer interrupt received exactly once for mask irq test
PASS: sbi: time: set_timer: timer immediately pending by setting timer to 0
PASS: sbi: time: set_timer: pending timer cleared while masked
PASS: sbi: ipi: Bad FID: SBI_ERR_NOT_SUPPORTED
SKIP: sbi: ipi: At least 2 cpus required
PASS: sbi: hsm: Bad FID: SBI_ERR_NOT_SUPPORTED
PASS: sbi: hsm: hart_get_status: status of current hart is started
SKIP: sbi: hsm: no other cpus to run the remaining hsm tests on
PASS: sbi: dbcn: Bad FID: SBI_ERR_NOT_SUPPORTED
DBCN_WRITE_TEST_STRING
PASS: sbi: dbcn: write: write success (error=0)
INFO: sbi: dbcn: write: 1 sbi calls made
DBCN_WRITE_TEST_STRING
PASS: sbi: dbcn: write: page boundary: write success (error=0)
INFO: sbi: dbcn: write: page boundary: 1 sbi calls made
SKIP: sbi: dbcn: write: high boundary: Memory above 4G required
SKIP: sbi: dbcn: write: high page: Memory above 4G required
SKIP: sbi: dbcn: write: invalid parameter: missing INVALID_ADDR environment variable
DBCN_WRITE_BYTE TEST BYTE: a
PASS: sbi: dbcn: write_byte: write success (error=0)
PASS: sbi: dbcn: write_byte: expected ret.value (0)
DBCN_WRITE_BYTE TEST WORD: a
PASS: sbi: dbcn: write_byte: write success (error=0)
PASS: sbi: dbcn: write_byte: expected ret.value (0)
PASS: sbi: susp: Bad FID: SBI_ERR_NOT_SUPPORTED
PASS: sbi: susp: funcid != 0 not supported
PASS: sbi: susp: basic: suspend and resume
PASS: sbi: susp: sleep_type: got expected sbi.error (-3)
PASS: sbi: susp: sleep_type upper bits: suspend and resume
SKIP: sbi: susp: bad addr: missing INVALID_ADDR environment variable
SKIP: sbi: susp: one cpu online: At least 2 cpus required
SKIP: sbi: sse: extension not available
PASS: sbi: fwft: Bad FID: SBI_ERR_NOT_SUPPORTED
PASS: sbi: fwft: base: get reserved feature 0x6: SBI_ERR_DENIED
PASS: sbi: fwft: base: set reserved feature 0x6: SBI_ERR_DENIED
PASS: sbi: fwft: base: get reserved feature 0x3fffffff: SBI_ERR_DENIED
PASS: sbi: fwft: base: set reserved feature 0x3fffffff: SBI_ERR_DENIED
PASS: sbi: fwft: base: get reserved feature 0x80000000: SBI_ERR_DENIED
PASS: sbi: fwft: base: set reserved feature 0x80000000: SBI_ERR_DENIED
PASS: sbi: fwft: base: get reserved feature 0xbfffffff: SBI_ERR_DENIED
PASS: sbi: fwft: base: set reserved feature 0xbfffffff: SBI_ERR_DENIED
PASS: sbi: fwft: misaligned_exc_deleg: Get misaligned deleg feature: SBI_SUCCESS
SKIP: sbi: fwft: misaligned_exc_deleg: missing MISALIGNED_EXC_DELEG_RESET environment variable
PASS: sbi: fwft: misaligned_exc_deleg: Set misaligned deleg feature invalid value 2: SBI_ERR_INVALID_PARAM
PASS: sbi: fwft: misaligned_exc_deleg: Set misaligned deleg feature invalid value 0xFFFFFFFF: SBI_ERR_INVALID_PARAM
PASS: sbi: fwft: misaligned_exc_deleg: Set misaligned deleg with invalid value > 32bits: SBI_ERR_INVALID_PARAM
PASS: sbi: fwft: misaligned_exc_deleg: Set misaligned deleg with invalid flag > 32bits: SBI_ERR_INVALID_PARAM
PASS: sbi: fwft: misaligned_exc_deleg: set to 0: SBI_SUCCESS
PASS: sbi: fwft: misaligned_exc_deleg: get 0 after set
PASS: sbi: fwft: misaligned_exc_deleg: set to 1: SBI_SUCCESS
PASS: sbi: fwft: misaligned_exc_deleg: get 1 after set
SKIP: sbi: fwft: misaligned_exc_deleg: Misaligned load exception does not trap in S-mode
PASS: sbi: fwft: misaligned_exc_deleg: Set misaligned deleg feature value 0 and lock: SBI_SUCCESS
PASS: sbi: fwft: misaligned_exc_deleg: locked: Set to 0 without lock flag: SBI_ERR_DENIED_LOCKED
PASS: sbi: fwft: misaligned_exc_deleg: locked: Set to 0 with lock flag: SBI_ERR_DENIED_LOCKED
PASS: sbi: fwft: misaligned_exc_deleg: locked: Set to 1 without lock flag: SBI_ERR_DENIED_LOCKED
PASS: sbi: fwft: misaligned_exc_deleg: locked: Set to 1 with lock flag: SBI_ERR_DENIED_LOCKED
PASS: sbi: fwft: misaligned_exc_deleg: locked: Get value 0
SKIP: sbi: fwft: pte_ad_hw_updating: not supported by platform
SKIP: sbi: fwft: dbtr: extension not available
SUMMARY: 74 tests, 18 skipped

EXIT: STATUS=1
PASS sbi (74 tests, 18 skipped)
```

inochisa:
Hi, @Dolu1990 , now we have linux in kvm virtual machine

Command 
```bash
qemu-system-riscv64 \
  -nographic --enable-kvm -machine virt \
  -m 512M -smp cpus=1 -cpu host \
  -kernel /root/Image \
  -initrd /root/initramfs.cpio \
  -append "console=ttyS0,115200n8 earlycon ignore_loglevel"
```

Boot log
```log
[    0.000000] Booting Linux on hartid 0
[    0.000000] Linux version 7.0.0-rc3-dirty (riscv64-linux-gnu-gcc (GCC) 15.1.0, GNU ld (GNU Binutils) 2.44) #48 SMP PREEMPT Wed Mar 25 10:53:17 CST 2026
[    0.000000] random: crng init done
[    0.000000] Machine model: riscv-virtio,qemu
[    0.000000] SBI specification v3.0 detected
[    0.000000] SBI implementation ID=0x3 Version=0x70000
[    0.000000] SBI TIME extension detected
[    0.000000] SBI IPI extension detected
[    0.000000] SBI RFENCE extension detected
[    0.000000] SBI SRST extension detected
[    0.000000] SBI DBCN extension detected
[    0.000000] SBI FWFT extension detected
[    0.000000] printk: debug: ignoring loglevel setting.
[    0.000000] efi: UEFI not found.
[    0.000000] earlycon: ns16550a0 at MMIO 0x0000000010000000 (options '')
[    0.000000] printk: legacy bootconsole [ns16550a0] enabled
[    0.000000] OF: reserved mem: Reserved memory: No reserved-memory node in the DT
[    0.000000] SBI HSM extension detected
[    0.000000] riscv: base ISA extensions acdfim
[    0.000000] riscv: ELF capabilities acdfim
[    0.000000] Ticket spinlock: enabled
[    0.000000] Zone ranges:
[    0.000000]   DMA32    [mem 0x0000000080000000-0x000000009fffffff]
[    0.000000]   Normal   empty
[    0.000000] Movable zone start for each node
[    0.000000] Early memory node ranges
[    0.000000]   node   0: [mem 0x0000000080000000-0x000000009fffffff]
[    0.000000] Initmem setup node 0 [mem 0x0000000080000000-0x000000009fffffff]
[    0.000000] percpu: Embedded 31 pages/cpu s88280 r8192 d30504 u126976
[    0.000000] pcpu-alloc: s88280 r8192 d30504 u126976 alloc=31*4096
[    0.000000] pcpu-alloc: [0] 0 
[    0.000000] Kernel command line: console=ttyS0,115200n8 earlycon ignore_loglevel
[    0.000000] printk: log buffer data + meta data: 131072 + 458752 = 589824 bytes
[    0.000000] Dentry cache hash table entries: 65536 (order: 7, 524288 bytes, linear)
[    0.000000] Inode-cache hash table entries: 32768 (order: 6, 262144 bytes, linear)
[    0.000000] Built 1 zonelists, mobility grouping on.  Total pages: 131072
[    0.000000] mem auto-init: stack:all(zero), heap alloc:off, heap free:off
[    0.000000] SLUB: HWalign=64, Order=0-3, MinObjects=0, CPUs=1, Nodes=1
[    0.000000] rcu: Hierarchical RCU implementation.
[    0.000000] rcu:     RCU event tracing is enabled.
[    0.000000] rcu:     RCU restricting CPUs from NR_CPUS=64 to nr_cpu_ids=1.
[    0.000000]  Trampoline variant of Tasks RCU enabled.
[    0.000000]  Tracing variant of Tasks RCU enabled.
[    0.000000] rcu: RCU calculated value of scheduler-enlistment delay is 25 jiffies.
[    0.000000] rcu: Adjusting geometry for rcu_fanout_leaf=16, nr_cpu_ids=1
[    0.000000] RCU Tasks: Setting shift to 0 and lim to 1 rcu_task_cb_adjust=1 rcu_task_cpu_ids=1.
[    0.000000] NR_IRQS: 64, nr_irqs: 64, preallocated irqs: 0
[    0.000000] riscv-intc: 64 local interrupts mapped
[    0.000000] riscv: providing IPIs using SBI IPI extension
[    0.000000] rcu: srcu_init: Setting srcu_struct sizes based on contention.
[    0.000000] clocksource: riscv_clocksource: mask: 0xffffffffffffffff max_cycles: 0x171024e7e0, max_idle_ns: 440795205315 ns
[    0.000041] sched_clock: 64 bits at 100MHz, resolution 10ns, wraps every 4398046511100ns
[    0.615826] Console: colour dummy device 80x25
[    0.931881] Calibrating delay loop (skipped), value calculated using timer frequency.. 200.00 BogoMIPS (lpj=400000)
[    1.676157] pid_max: default: 32768 minimum: 301
[    2.069036] Mount-cache hash table entries: 1024 (order: 1, 8192 bytes, linear)
[    2.573399] Mountpoint-cache hash table entries: 1024 (order: 1, 8192 bytes, linear)
[    3.255621] VFS: Finished mounting rootfs on nullfs
[    4.047842] riscv: ELF compat mode unsupported
[    4.048055] ASID allocator disabled (0 bits)
[    4.688486] rcu: Hierarchical SRCU implementation.
[    5.020970] rcu:     Max phase no-delay instances is 1000.
[    5.467583] EFI services will not be available.
[    5.804331] smp: Bringing up secondary CPUs ...
[    6.125065] smp: Brought up 1 node, 1 CPU
[    6.421377] Memory: 424740K/524288K available (11777K kernel code, 5982K rwdata, 10240K rodata, 2439K init, 389K bss, 96548K reserved, 0K cma-reserved)
[    7.437406] devtmpfs: initialized
[    7.896456] clocksource: jiffies: mask: 0xffffffff max_cycles: 0xffffffff, max_idle_ns: 7645041785100000 ns
[    8.567706] posixtimers hash table entries: 512 (order: 1, 8192 bytes, linear)
[    9.103648] futex hash table entries: 256 (16384 bytes on 1 NUMA nodes, total 16 KiB, linear).
[    9.852842] DMI not present or invalid.
[   10.205903] NET: Registered PF_NETLINK/PF_ROUTE protocol family
[   10.657680] DMA: preallocated 128 KiB GFP_KERNEL|GFP_DMA32 pool for atomic allocations
[   11.237146] audit: initializing netlink subsys (disabled)
[   11.659839] audit: type=2000 audit(0.748:1): state=initialized audit_enabled=0 res=1
[   12.273113] thermal_sys: Registered thermal governor 'step_wise'
[   12.280300] cpuidle: using governor menu
[   13.028902] SBI misaligned access exception delegation ok
[   13.888321] HugeTLB: registered 2.00 MiB page size, pre-allocated 0 pages
[   14.863779] HugeTLB: 28 KiB vmemmap can be freed for a 2.00 MiB page
[   16.061318] ACPI: Interpreter disabled.
[   16.629115] iommu: Default domain type: Translated
[   16.970005] iommu: DMA domain TLB invalidation policy: strict mode
[   18.312389] SCSI subsystem initialized
[   18.874083] libata version 3.00 loaded.
[   19.497546] usbcore: registered new interface driver usbfs
[   20.313994] usbcore: registered new interface driver hub
[   20.677283] usbcore: registered new device driver usb
[   21.961094] Advanced Linux Sound Architecture Driver Initialized.
[   23.169633] vgaarb: loaded
[   23.582227] clocksource: Switched to clocksource riscv_clocksource
[   24.957778] pnp: PnP ACPI: disabled
[   27.460399] NET: Registered PF_INET protocol family
[   29.124482] IP idents hash table entries: 8192 (order: 4, 65536 bytes, linear)
[   34.957448] tcp_listen_portaddr_hash hash table entries: 256 (order: 0, 4096 bytes, linear)
[   35.636269] Table-perturb hash table entries: 65536 (order: 6, 262144 bytes, linear)
[   35.953513] TCP established hash table entries: 4096 (order: 3, 32768 bytes, linear)
[   36.889176] TCP bind hash table entries: 4096 (order: 5, 131072 bytes, linear)
[   37.511411] TCP: Hash tables configured (established 4096 bind 4096)
[   38.048264] UDP hash table entries: 256 (order: 2, 16384 bytes, linear)
[   39.150022] UDP-Lite hash table entries: 256 (order: 2, 16384 bytes, linear)
[   40.409102] NET: Registered PF_UNIX/PF_LOCAL protocol family
[   41.368940] RPC: Registered named UNIX socket transport module.
[   42.232025] RPC: Registered udp transport module.
[   42.576651] RPC: Registered tcp transport module.
[   43.672180] RPC: Registered tcp-with-tls transport module.
[   44.064092] RPC: Registered tcp NFSv4.1 backchannel transport module.
[   45.559946] PCI: CLS 0 bytes, default 64
[   45.880960] kvm [1]: hypervisor extension not available
[   46.116329] Unpacking initramfs...
[   46.684292] workingset: timestamp_bits=46 max_order=17 bucket_order=0
[   47.649144] NFS: Registering the id_resolver key type
[   48.235915] Key type id_resolver registered
[   48.400784] Key type id_legacy registered
[   49.212755] nfs4filelayout_init: NFSv4 File Layout Driver Registering...
[   50.144314] nfs4flexfilelayout_init: NFSv4 Flexfile Layout Driver Registering...
[   51.083896] 9p: Installing v9fs 9p2000 file system support
[   51.764965] Block layer SCSI generic (bsg) driver version 0.4 loaded (major 244)
[   52.736528] io scheduler mq-deadline registered
[   52.921565] io scheduler kyber registered
[   53.803327] io scheduler bfq registered
[   54.743506] riscv-plic: plic@c000000: mapped 96 interrupts with 1 handlers for 1 contexts.
[   55.899943] pci-host-generic 30000000.pci: host bridge /soc/pci@30000000 ranges:
[   56.864249] pci-host-generic 30000000.pci:       IO 0x0003000000..0x000300ffff -> 0x0000000000
[   58.788218] pci-host-generic 30000000.pci:      MEM 0x0040000000..0x007fffffff -> 0x0040000000
[   59.132185] pci-host-generic 30000000.pci:      MEM 0x0400000000..0x07ffffffff -> 0x0400000000
[   60.843596] pci-host-generic 30000000.pci: Memory resource size exceeds max for 32 bits
[   61.841066] pci-host-generic 30000000.pci: ECAM at [mem 0x30000000-0x3fffffff] for [bus 00-ff]
[   62.984596] pci-host-generic 30000000.pci: PCI host bridge to bus 0000:00
[   63.840571] pci_bus 0000:00: root bus resource [bus 00-ff]
[   64.064693] pci_bus 0000:00: root bus resource [io  0x0000-0xffff]
[   65.291642] pci_bus 0000:00: root bus resource [mem 0x40000000-0x7fffffff]
[   65.568760] pci_bus 0000:00: root bus resource [mem 0x400000000-0x7ffffffff]
[   67.008938] pci 0000:00:00.0: [1b36:0008] type 00 class 0x060000 conventional PCI endpoint
[   68.252824] pci_bus 0000:00: resource 4 [io  0x0000-0xffff]
[   68.484502] pci_bus 0000:00: resource 5 [mem 0x40000000-0x7fffffff]
[   69.715622] pci_bus 0000:00: resource 6 [mem 0x400000000-0x7ffffffff]
[   77.168602] Serial: 8250/16550 driver, 4 ports, IRQ sharing disabled
[   78.220759] printk: legacy console [ttyS0] disabled
[   78.876294] 10000000.serial: ttyS0 at MMIO 0x10000000 (irq = 12, base_baud = 230400) is a 16550A
[   79.993566] printk: legacy console [ttyS0] enabled
[   79.993566] printk: legacy console [ttyS0] enabled
[   81.183521] printk: legacy bootconsole [ns16550a0] disabled
[   81.183521] printk: legacy bootconsole [ns16550a0] disabled
[   83.813724] loop: module loaded
[   84.685123] e1000e: Intel(R) PRO/1000 Network Driver
[   85.307480] e1000e: Copyright(c) 1999 - 2015 Intel Corporation.
[   86.153988] usbcore: registered new interface driver uas
[   86.851950] usbcore: registered new interface driver usb-storage
[   87.603966] mousedev: PS/2 mouse device common for all mice
[   88.476329] goldfish_rtc 101000.rtc: registered as rtc0
[   89.156799] goldfish_rtc 101000.rtc: setting system clock to 2026-02-10T04:06:11 UTC (1770696371)
[   90.385302] sdhci: Secure Digital Host Controller Interface driver
[   91.155538] sdhci: Copyright(c) Pierre Ossman
[   91.716300] Synopsys Designware Multimedia Card Interface Driver
[   92.477482] sdhci-pltfm: SDHCI platform and OF driver helper
[   93.332083] usbcore: registered new interface driver usbhid
[   94.051494] usbhid: USB HID core driver
[   94.631517] NET: Registered PF_INET6 protocol family
[   95.433896] Segment Routing with IPv6
[   95.916121] In-situ OAM (IOAM) with IPv6
[   96.409049] sit: IPv6, IPv4 and MPLS over IPv4 tunneling driver
[   97.232274] NET: Registered PF_PACKET protocol family
[   97.853961] 9pnet: Installing 9P2000 support
[   98.408778] Key type dns_resolver registered
[  102.009619] Legacy PMU implementation is available
[  109.665507] clk: Disabling unused clocks
[  109.829305] PM: genpd: Disabling unused power domains
[  110.435620] ALSA device list:
[  110.553265]   No soundcards found.
[  114.201551] Freeing initrd memory: 38256K
[  114.659325] Freeing unused kernel image (initmem) memory: 2436K
[  114.857756] Run /init as init process
[  114.979664]   with arguments:
[  115.081629]     /init
[  115.156619]   with environment:
[  115.267657]     HOME=/
[  115.344899]     TERM=linux
Starting syslogd: OK
Running sysctl: OK
Populating /dev using udev: [  118.863881] udevd[72]: starting version 3.2.14
[  119.369982] udevd[73]: starting eudev-3.2.14
done
Starting crond: OK

test login: root
[root@test /]# coremark 0x0 0x0 0x66 0 7 1 2000
2K performance run parameters for coremark.
CoreMark Size    : 666
Total ticks      : 8127
Total time (secs): 8.127000
Iterations/Sec   : 135.351298
ERROR! Must execute for at least 10 secs for a valid result!
Iterations       : 1100
Compiler version : GCC14.2.0
Compiler flags   : -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -O2 -g0    -lrt
Memory location  : Please put data memory location here
                        (e.g. code in flash, data on heap etc)
seedcrc          : 0xe9f5
[0]crclist       : 0xe714
[0]crcmatrix     : 0x1fd7
[0]crcstate      : 0x8e3a
[0]crcfinal      : 0x33ff
Errors detected
[root@test /]# coremark 0x3415 0x3415 0x66 0 7 1 2000
2K validation run parameters for coremark.
CoreMark Size    : 666
Total ticks      : 8063
Total time (secs): 8.063000
Iterations/Sec   : 136.425648
ERROR! Must execute for at least 10 secs for a valid result!
Iterations       : 1100
Compiler version : GCC14.2.0
Compiler flags   : -D_LARGEFILE_SOURCE -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64  -O2 -g0    -lrt
Memory location  : Please put data memory location here
                        (e.g. code in flash, data on heap etc)
seedcrc          : 0x18f2
[0]crclist       : 0xe3c1
[0]crcmatrix     : 0x0747
[0]crcstate      : 0x8d84
[0]crcfinal      : 0x30f5
Errors detected
[root@test /]# cat /proc/cpuinfo 
processor       : 0
hart            : 0
isa             : rv64imafdc_zicntr_zicsr_zifencei_zihpm_zaamo_zalrsc_zca_zcd
mmu             : sv39
mvendorid       : 0x0
marchid         : 0x2e
mimpid          : 0x0
hart isa        : rv64imafdc_zicntr_zicsr_zifencei_zihpm_zaamo_zalrsc_zca_zcd
```

inochisa:
I have add a simple doc for this implementation at SpinalHDL/VexiiRiscv-RTD#14

Dolu1990:
Hi @inochisa ,

> Linux could be a problem, not for a synthesized core, but for the simulation. In fact, I managed to boot an mainline OpenSBI with the simulation. And it gave me a 1.1GB trace log and a 5.1 GB fst wave. XD

Right ^^
The solution to that is to use the --dual-sim option in addition.
This will start  2 identitcal simulation without wave with a delay in between of about 1Mcycle.
When the simulation ahead crash, this will enable the wave on the second simulation for that last 1Mcycle.
That way you have the wave were it is about interresting. In other words, the first simulation is used as a trigger for the second simulation to start capture the wave.

Else in general, not using --with-wave at all (and without others similar options aswell)

> And at least for now, my team are struggled to boot a linux with rootfs. Since my board has no SD card and I have to boot the kernel with ethernet.

How big is the kernel binary and rootfs ?
(For buildroot, i generaly boot using jtag, when it is boosted, it goes about  800Kbits/s, so if the images aren't tooooo big, that is ok. but probably to slow for your images

> My suggestion is adding a batch of unit tests like defermelowie/riscv-hext-asm-tests (And I just use this to test the implementation).

Why not :)

> (By this way, I think the spike for Vexiiriscv is too old to support H extension. XD)

Did you tried merging upstream ?
I did a few change in the fork, hopefully not toooo much.

> With lots of unit test are required, I think a small framework and some common code are needed to invoke the unit test automatically

So far, in vexii, RegressionSingle does that. but it isn't a very nice code base XD

> And a good news, The kvm unit test has passed.

Nice ^^

> now we have linux in kvm virtual machine

So, to be sure i understand : 
- You have a FPGA board running VexiiRiscv + linux, in which you run a kvm virtual machine running linux (?!?)

> I have add a simple doc for this implementation at SpinalHDL/VexiiRiscv-RTD#14

Thanks :D

inochisa:
Hi @Dolu1990,

> Hi @inochisa ,
> 
> > Linux could be a problem, not for a synthesized core, but for the simulation. In fact, I managed to boot an mainline OpenSBI with the simulation. And it gave me a 1.1GB trace log and a 5.1 GB fst wave. XD
> 
> Right ^^ The solution to that is to use the --dual-sim option in addition. This will start 2 identitcal simulation without wave with a delay in between of about 1Mcycle. When the simulation ahead crash, this will enable the wave on the second simulation for that last 1Mcycle. That way you have the wave were it is about interresting. In other words, the first simulation is used as a trigger for the second simulation to start capture the wave.
> 
> Else in general, not using --with-wave at all (and without others similar options aswell)
> 

That's a good news. And after the test for a  synthesized core. If there is already a way to boot  OpenSBI and linux without hypervisor. It can also boot the linux with H extension, just set KVM is y. And I guess you already has a for it?

> > And at least for now, my team are struggled to boot a linux with rootfs. Since my board has no SD card and I have to boot the kernel with ethernet.
> 
> How big is the kernel binary and rootfs ? (For buildroot, i generaly boot using jtag, when it is boosted, it goes about 800Kbits/s, so if the images aren't tooooo big, that is ok. but probably to slow for your images
> 

Oh, We have boot with another board with SD card. And we have a pretty big Image and rootfs. Because all the thing are build from upsteam (Linux is build from the defconfig and the initramfs is build from a full-featured buildroot rootfs (about 45Mib without compress))

> > My suggestion is adding a batch of unit tests like defermelowie/riscv-hext-asm-tests (And I just use this to test the implementation).
> 
> Why not :)
> 

Good, I will try to add it. But it will be a long fight XD.

> > (By this way, I think the spike for Vexiiriscv is too old to support H extension. XD)
> 
> Did you tried merging upstream ? I did a few change in the fork, hopefully not toooo much.
> 

The last time I have tried to upgrade is a bad time. Too many compile error when upgrading the spike to the master. The spike is bad for not giving a new release for a long time, which make me hard to figure out the change that breaks API.... 

> > With lots of unit test are required, I think a small framework and some common code are needed to invoke the unit test automatically
> 
> So far, in vexii, RegressionSingle does that. but it isn't a very nice code base XD
> 

I will take a look. Thanks.

> > And a good news, The kvm unit test has passed.
> 
> Nice ^^
> 
> > now we have linux in kvm virtual machine
> 
> So, to be sure i understand :
> 
>     * You have a FPGA board running VexiiRiscv + linux, in which you run a kvm virtual machine running linux (?!?)
> 

Right. And all the softwares (linux, OpenSBI, rootfs) are using upstream version without any changes

> 
> > I have add a simple doc for this implementation at [SpinalHDL/VexiiRiscv-RTD#14](https://github.com/SpinalHDL/VexiiRiscv-RTD/pull/14)
> 
> Thanks :D

Dolu1990:
>  It can also boot the linux with H extension, just set KVM is y. 

Ahhh nice ^^
So far, i was using https://github.com/SpinalHDL/NaxSoftware/tree/ae0c08016d0581b9ac79a4feb166e7a532369bd4/buildroot
to build the images, but it is quite outdated.

> And I guess you already has a for it?

has a (?) for it ? (i don't understand :D)

> The last time I have tried to upgrade is a bad time. Too many compile error when upgrading the spike to the master. The spike is bad for not giving a new release for a long time, which make me hard to figure out the change that breaks API....

In case, let me know, there is maybe some change i did which can be reverted. In particular, the one related to the MMU.

> Right. And all the softwares (linux, OpenSBI, rootfs) are using upstream version without any changes

GG :D :D

inochisa:
> > It can also boot the linux with H extension, just set KVM is y.
> 
> Ahhh nice ^^ So far, i was using https://github.com/SpinalHDL/NaxSoftware/tree/ae0c08016d0581b9ac79a4feb166e7a532369bd4/buildroot to build the images, but it is quite outdated.
> 

I just use the upstream buildroot. And it seems like no change is needed XD.

> > And I guess you already has a for it?
> 
> has a (?) for it ? (i don't understand :D)
> 

has a working Linux with the way of simulation you have mentioned (without hypervisor enabled).

> > The last time I have tried to upgrade is a bad time. Too many compile error when upgrading the spike to the master. The spike is bad for not giving a new release for a long time, which make me hard to figure out the change that breaks API....
> 
> In case, let me know, there is maybe some change i did which can be reverted. In particular, the one related to the MMU.
> 

Great, I will take a look for it.

> > Right. And all the softwares (linux, OpenSBI, rootfs) are using upstream version without any changes
> 
> GG :D :D

Dolu1990:
> has a working Linux with the way of simulation you have mentioned (without hypervisor enabled).

Ahh right, and it is integrated that way in RegressionSingle : 

https://github.com/SpinalHDL/VexiiRiscv/blob/b1597e52be94bf9d66aab091fec0fdc978d91fe3/src/test/scala/vexiiriscv/tester/RegressionSingle.scala#L395

Those various call are translated into regular [TestBench](https://github.com/SpinalHDL/VexiiRiscv/blob/b1597e52be94bf9d66aab091fec0fdc978d91fe3/src/main/scala/vexiiriscv/tester/TestBench.scala) arguments ^^

inochisa:
> > has a working Linux with the way of simulation you have mentioned (without hypervisor enabled).
> 
> Ahh right, and it is integrated that way in RegressionSingle :
> 
> https://github.com/SpinalHDL/VexiiRiscv/blob/b1597e52be94bf9d66aab091fec0fdc978d91fe3/src/test/scala/vexiiriscv/tester/RegressionSingle.scala#L395
> 
> Those various call are translated into regular [TestBench](https://github.com/SpinalHDL/VexiiRiscv/blob/b1597e52be94bf9d66aab091fec0fdc978d91fe3/src/main/scala/vexiiriscv/tester/TestBench.scala) arguments ^^

Great, so the only thing we need is setting the KVM is y. And adjust the DTB. And everything will be fine. Right?

Dolu1990:
> Great, so the only thing we need is setting the KVM is y. And adjust the DTB. And everything will be fine. Right?

- And adding the commands to boot linux in linux (for configs with rvh)
- double the timeout for configs with rvh
- Maybe the opensbi version / linux version used in NaxSoftware are too old

But yes, in a first time, probably just adding the KVM option to see if with the new images it passes is good.

inochisa:
> > Great, so the only thing we need is setting the KVM is y. And adjust the DTB. And everything will be fine. Right?
> 
>     * And adding the commands to boot linux in linux (for configs with rvh)

This is a problem, as booting linux lead a nested build, which is hard to handle.

> 
>     * double the timeout for configs with rvh
> 
>     * Maybe the opensbi version / linux version used in NaxSoftware are too old
> 

These two are fine and I can fix it.

> 
> But yes, in a first time, probably just adding the KVM option to see if with the new images it passes is good.

Thanks, I will update this branch for it. ^^

inochisa:
> > > Great, so the only thing we need is setting the KVM is y. And adjust the DTB. And everything will be fine. Right?
> > 
> > 
> > ```
> > * And adding the commands to boot linux in linux (for configs with rvh)
> > ```
> 
> This is a problem, as booting linux lead a nested build, which is hard to handle.
> 

Or we can add kvm-unit-test for a fast test? I think it is also a good alternative.

Dolu1990:
> Or we can add kvm-unit-test for a fast test? I think it is also a good alternative.

Thoes should be there anyway i would say.

The linux in linux test would be a test which would allow to exercise the CPU in all its potential cycle/timings bugs (direct tests often miss cycle/timing related bugs, which are horrible to debug on hardware)

> This is a problem, as booting linux lead a nested build, which is hard to handle.

Ahh i don't know much about it at all.
You mean that it is mostly a "hand crafted assembly" of two different builds ?

inochisa:
> > Or we can add kvm-unit-test for a fast test? I think it is also a good alternative.
> 
> Thoes should be there anyway i would say.
> 
> The linux in linux test would be a test which would allow to exercise the CPU in all its potential cycle/timings bugs (direct tests often miss cycle/timing related bugs, which are horrible to debug on hardware)
> 

Got it.

> > This is a problem, as booting linux lead a nested build, which is hard to handle.
> 
> Ahh i don't know much about it at all. You mean that it is mostly a "hand crafted assembly" of two different builds ?

The are two things for running a VM:
1. the kernel, we can just reuse the kernel for the simulator
2. the rootfs/initramfs

The rootfs for the guest can just use it for the host. In this way we need to make a copy and move this copy in the rootfs. This is what I mean the nested build. If the rootfs can adapt mulltiple console.

But the thing is kind of different. The initramfs for the guest kernel is more like a normal machine, so this means a different build is required. And It could be better to make a small initramfs for the guest so we can reduce the loading time when invoking the Qemu/kvmtools. This is the thing I would to address.

In my option, at least the following in the buildroot rootfs.
1. Qemu (it is widely used and we can simplify it a lot, and it is recommended by kvm-unit-test).
2. bash (required by kvm-unit-test)

Dolu1990:
> And It could be better to make a small initramfs for the guest so we can reduce the loading time when invoking the Qemu/kvmtools. This is the thing I would to address.

Ahhh right.
Sound good to me.

inochisa:
HI, @Dolu1990 
I am sure where these definitions  come from? Is there any doc for it?

https://github.com/SpinalHDL/NaxSoftware/blob/main/buildroot/boards/naxriscv_sim/linux/linux_raw.config#L5-L8

Dolu1990:
> HI, @Dolu1990 I am sure where these definitions come from? Is there any doc for it?
> 
> https://github.com/SpinalHDL/NaxSoftware/blob/main/buildroot/boards/naxriscv_sim/linux/linux_raw.config#L5-L8

Hi ^^

Those are custom things i added to support multiple configs. They are handled via that command bellow, which will underthehood use gcc preprocessor to generate the final defconfigs : 
https://github.com/SpinalHDL/NaxSoftware/blob/ae0c08016d0581b9ac79a4feb166e7a532369bd4/buildroot/README.md?plain=1#L10

(if i remember well)
rv32ima refers to https://github.com/SpinalHDL/NaxSoftware/blob/main/buildroot/configs/rv32ima.sh for instance.
So you can add new architecture by adding new .sh files there and refering to them when you run the https://github.com/SpinalHDL/NaxSoftware/blob/main/buildroot/naxriscv_sim_setup.sh scripts as shown in the https://github.com/SpinalHDL/NaxSoftware/blob/ae0c08016d0581b9ac79a4feb166e7a532369bd4/buildroot/README.md?plain=1#L10 example.

inochisa:
> > HI, @Dolu1990 I am sure where these definitions come from? Is there any doc for it?
> > https://github.com/SpinalHDL/NaxSoftware/blob/main/buildroot/boards/naxriscv_sim/linux/linux_raw.config#L5-L8
> 
> Hi ^^
> 
> Those are custom things i added to support multiple configs. They are handled via that command bellow, which will underthehood use gcc preprocessor to generate the final defconfigs : https://github.com/SpinalHDL/NaxSoftware/blob/ae0c08016d0581b9ac79a4feb166e7a532369bd4/buildroot/README.md?plain=1#L10
> 
> (if i remember well) rv32ima refers to https://github.com/SpinalHDL/NaxSoftware/blob/main/buildroot/configs/rv32ima.sh for instance. So you can add new architecture by adding new .sh files there and refering to them when you run the https://github.com/SpinalHDL/NaxSoftware/blob/main/buildroot/naxriscv_sim_setup.sh scripts as shown in the https://github.com/SpinalHDL/NaxSoftware/blob/ae0c08016d0581b9ac79a4feb166e7a532369bd4/buildroot/README.md?plain=1#L10 example.

Thanks, I already found some of them, but I missed some script. And I think it is great for me to update them for this PR.

inochisa:
Update: multiple fixes that fix some case condition that will cause the CPU broken. Also a small optimization for better performance. For test, the score of coremark is 138, a small improvement, but it is stable now (The old result is not stable and usually had a bad score about 90~130).

inochisa:
> Note, coremark is quite sensitive to LSU latency, as it does quite a bit of linked list pointer chassing.

An interesting thing is, when build H extension with SSTC extension, the score of coremark has increased dramatically to 159 (and the host part is to 191). So the LSU latency is one of burden, but not all the things.

I guess this worth a note for others that wants this extension.

And the performance loss is less than 20% in this case, which is quite acceptable I guess. I have test my dev board with a latest RVA23 CPU (H + AIA + SSTC), it has about 9.7% performance loss for VS-mode (at least for coremark)

Dolu1990:
> multiple fixes that fix some case condition that will cause the CPU broken

Related to change in this PR ? or pre-existing bugs ? (To know if i need to notify other peoples ^^)

> An interesting thing is, when build H extension with SSTC extension, the score of coremark has increased dramatically to 159 (and the host part is to 191). So the LSU latency is one of burden, but not all the things.

At which frequency are you running the CPU ? which board ? 
On designs which run at <50 MHz, or with slow memory system, i noticed that the linux default tick rate (250 hz) + usage of the opensbi timer start to make the CPU busy a large % of time.

inochisa:
> > multiple fixes that fix some case condition that will cause the CPU broken
> 
> Related to change in this PR ? or pre-existing bugs ? (To know if i need to notify other peoples ^^)
> 

Yes, all the fixes are only related to this PR if I filter them correctly. Changes not related to this PR are already splitted (like recent two new PRs). I will rebase this branch once the local build is passed.

> > An interesting thing is, when build H extension with SSTC extension, the score of coremark has increased dramatically to 159 (and the host part is to 191). So the LSU latency is one of burden, but not all the things.
> 
> At which frequency are you running the CPU ? which board ? 

It is at 100Mhz. Our board is not official supported by the litex, you can find the board file here https://github.com/looongbin/litex-boards/commit/2258ff3db49c13c1375dc8780897cf97f923fe7a. It is an ordinary XCKU5P board and nothing special.

> On designs which run at <50 MHz, or with slow memory system, i noticed that the linux default tick rate (250 mhz) + usage of the opensbi timer start to make the CPU busy a large % of time.

Yeah, it is the reason, SSTC avoid the time trap to M-mode for S-mode and the time trap to HS-mode for VS-mode. This saves a lot of CPU time as the timer interrupt is frequent. In fact I am not very surprised about this, but I think it is a good thing to share. XD.

inochisa:
The linux tick uses the fixed time interval for timer interrupt, which means low frequency CPU needs more CPU time (in percentage) for timer interrupt.

inochisa:
A small update for hypervisor Zicbom support.

inochisa:
@Dolu1990  ~A bad news, I have failed to porting the latest riscv-arch-test for VexiiRiscv. The test now requires a 16550 UART.  XD~ 
~Just need to porting the IO command, I am working on it~
Bad news again, there is no hypervisor support for sail riscv, but I have ported to the latest riscv-arch-test successfully.

And I guess it could be better to use a real CLINT or PLIC for the TestBench?

Dolu1990:
Hi,

> And I guess it could be better to use a real CLINT or PLIC for the TestBench?

Ahhh the idea was to keep the testbench as light as possible and to emulate everything around.
Maybe your proposal could be done by implementing a new plugin which would intercept the IO bus and create the CLINT PLIC directly inside the testbench ?

Or else, putting in place a little simulation SoC.

inochisa:
> Hi,
> 
> > And I guess it could be better to use a real CLINT or PLIC for the TestBench?
> 
> Ahhh the idea was to keep the testbench as light as possible and to emulate everything around. 

I agree. This is necessary to make the simulation fast.

> Maybe your proposal could be done by implementing a new plugin which would intercept the IO bus and create the CLINT PLIC directly inside the testbench ?
> 
> Or else, putting in place a little simulation SoC.

Yeah, at least for now I think nothing needed. I have misunderstand the requirement for the "riscv-arch-test", and in fact it does not require a real PLIC/CLINT. Anyway, I am kind of upset as no hypervisor support for it. (I have told someone is doing this at the last RISC-V sig-hypervisor meeting, but I am sure it will be a long story to finish it XD)

inochisa:
Hi @Dolu1990, I have bumped the NaxSoftware, so I think this is also OK to be merged. And I can created PR for H-AIA support and some other improvements

mrcmry:
This is a bit unrelated, but "sbt:VexiiRiscv> Test/runMain vexiiriscv.tester.TestBench --with-mul --with-div --load-elf ext/NaxSoftware/baremetal/dhrystone/build/rv32ima/dhrystone.elf --trace-all" fail with the current dev and spinal 1.14.2 (also other previous version).

I mention this here because it seems related to IMSIC and I think it's part of H extension and your work @inochisa :

```
[info] compiling 140 Scala sources and 1 Java source to .../VexiiRiscv/target/scala-2.12/classes ...
[error].../VexiiRiscv/src/main/scala/vexiiriscv/misc/PrivilegedPlugin.scala:849:49: type mismatch;
[error]  found   : scala.collection.immutable.Range
[error]  required: Int
[error]         val file = ImsicFile(hartIds(hartId), 1 until p.imsicInterrupts)
[error]                                                 ^
[error] .../VexiiRiscv/src/main/scala/vexiiriscv/soc/TilelinkVexiiRiscvFiber.scala:96:43: overloaded method value := with alternatives:
[error]   (value: String)Unit <and>
[error]   (rangesValue: (Any, Any),_rangesValues: (Any, Any)*)Unit <and>
[error]   (that: spinal.core.Bits)(implicit loc: spinal.idslplugin.Location)Unit
[error]  cannot be applied to (spinal.lib.Stream[spinal.core.UInt])
[error]         case PrivilegeMode.M => priv.mmsi := msi.addImsicFileinfo(ImsicFileInfo(intIdBase, 1 until intNum))
[error] ...
```

inochisa:
> This is a bit unrelated, but "sbt:VexiiRiscv> Test/runMain vexiiriscv.tester.TestBench --with-mul --with-div --load-elf ext/NaxSoftware/baremetal/dhrystone/build/rv32ima/dhrystone.elf --trace-all" fail with the current dev and spinal 1.14.2 (also other previous version).
> 
> I mention this here because it seems related to IMSIC and I think it's part of H extension and your work @inochisa :
> 
> ```
> [info] compiling 140 Scala sources and 1 Java source to .../VexiiRiscv/target/scala-2.12/classes ...
> [error].../VexiiRiscv/src/main/scala/vexiiriscv/misc/PrivilegedPlugin.scala:849:49: type mismatch;
> [error]  found   : scala.collection.immutable.Range
> [error]  required: Int
> [error]         val file = ImsicFile(hartIds(hartId), 1 until p.imsicInterrupts)
> [error]                                                 ^
> [error] .../VexiiRiscv/src/main/scala/vexiiriscv/soc/TilelinkVexiiRiscvFiber.scala:96:43: overloaded method value := with alternatives:
> [error]   (value: String)Unit <and>
> [error]   (rangesValue: (Any, Any),_rangesValues: (Any, Any)*)Unit <and>
> [error]   (that: spinal.core.Bits)(implicit loc: spinal.idslplugin.Location)Unit
> [error]  cannot be applied to (spinal.lib.Stream[spinal.core.UInt])
> [error]         case PrivilegeMode.M => priv.mmsi := msi.addImsicFileinfo(ImsicFileInfo(intIdBase, 1 until intNum))
> [error] ...
> ```

@mrcmry I guess you bump the SpinalHDL to the latest? SpinalHDL/SpinalHDL#1914 change the interface of the ImsicFile. I have prepared all the change in another tree, and will send a PR when this is merged. Could you try this one? https://github.com/project-inochi/VexiiRiscv/tree/h-aia

mrcmry:
@inochisa it works mostly, but there is still a build error. I documented it in #133 (should have done this in the first place instead of polluting this PR discussion, sorry).

By the way @Dolu1990 , it would be useful to have at least one tag, so `git discribe --tags` works (I used the `git describe --all --long` as a workaround). Could be v0.0.0 of the first commit for example.

inochisa:
Hi @Dolu1990, As mentioned before, since there is a build problem, should I merge the change for AIA support with this PR?
I am not sure it is suitable as that support is kind of complex....

Dolu1990:
Hi,

@mrcmry  v0.0.0 added ^^
@inochisa I'm not sure to understand, with the gitmodule specified in the git repo it all build right ?

inochisa:
> @inochisa I'm not sure to understand, with the gitmodule specified in the git repo it all build right ?

@Dolu1990 Actually the reference in git repo is fine. In fact, if it has something wrong, the github check will not pass ^^

Dolu1990:
Hi,

I'm testing the PR using vexiiriscv.tester.Regression.
This utility generate random VexiiRiscv configuration and then run all the regressions test on it.

Currently, most generated configuration do not pass the SpinalHDL generation. (60% of broken config)

Here is how you can recreate the issue : 
- Merge VexiiRiscv upstream dev (just pushed 2 commit to add VEXIIRISCV_REGRESSION_SIM support)
- Run : 

```shell
export VEXIIRISCV_REGRESSION_SIM=0
export VEXIIRISCV_REGRESSION_THREAD_COUNT=22
sbt "testOnly vexiiriscv.tester.Regression"
```

All the failed configs will be stored in the regression/FAIL_xxxx files so you can pick one and test individualy via sbt "runMain vexiiriscv.Generate Here_content_of_FAIL_xxxx_file"

Once all of those passes, we can try : 

```shell
export VEXIIRISCV_REGRESSION_SIM=1
export VEXIIRISCV_REGRESSION_THREAD_COUNT=8
sbt "testOnly vexiiriscv.tester.Regression"
```

which will this time also run the simulations (here the paine will realy start :D :D)

So, the road map would be : 
- Get it to pass with export VEXIIRISCV_REGRESSION_SIM=0
- Then get it to pass with export VEXIIRISCV_REGRESSION_SIM=1

let me know if you have issues / questions
Also, it is fine to restrict RVH to a subset of configurations to avoid fixing configurations which make no practical sense

inochisa:
@Dolu1990, thanks for you detailed instruction, I have fixed all the build error and set up a full test to see if it can pass. I will call you again once the simulation success.

inochisa:
> let me know if you have issues / questions Also, it is fine to restrict RVH to a subset of configurations to avoid fixing configurations which make no practical sense

I guess it could be better and simple to restrict RVH with RVG, it is not very practical for a minimum configuration

Dolu1990:
> I guess it could be better and simple to restrict RVH with RVG, it is not very practical for a minimum configuration

Yes right
let me know if you need any help to setup things / how to debug

inochisa:
@Dolu1990, could you have a try, I have pass the regression test with the following parameters:
```bash
export VEXIIRISCV_REGRESSION_SIM=1
export VEXIIRISCV_REGRESSION_THREAD_COUNT=96
export VEXIIRISCV_REGRESSION_BUILDROOT_ENABLED=0
sbt "testOnly vexiiriscv.tester.Regression"
```
I have to disable the buildroot test as it cost too much time. XD
Last time I run the buildroot simulation (the thing in the NaxSoftware PR) leads to a run with more than three days in this tester....

Sorry for the truncated log as I forgot to adjust the zellij configuration.
[vexii.log](https://github.com/user-attachments/files/28431588/vexii.log)

Dolu1990:
Hi,

I just tried, and it fail on my end.

Looking at the logs you sent, i can see : 
[Warning] RVLS not detected

This mean the simulation are running, but aren't tested against the RVLS checker.
You probably need to compile ext/rvls.

See https://github.com/SpinalHDL/rvls/tree/94f850e5f7c9a36e5aceb3810d658fffae55b497#how-to-compile
=> 

```sh
# Compile riscv-isa-sim (spike), used as a golden model during the sim to check the dut behavior (lock-step)
cd VexiiRiscv/ext/riscv-isa-sim
mkdir build
cd build
# Optionally add CFLAGS='-g -O0' CXXFLAGS='-g -O0'
../configure --prefix=$RISCV --enable-commitlog  --without-boost --without-boost-asio --without-boost-regex
make -j$(nproc)
cd ../..

# Compile RVLS
cd rvls
make -j$(nproc)
```

Then, when you will run a simulation, as soon as there is a missmatch between the sim and the golden reference, it will stop the simulation and fail.
That is maybe why you had endless runtime with buildroot enable, maybe one simulation locked up, which would normaly have been detected by RVLS.

Good luck :D

inochisa:
> Hi,
> 
> I just tried, and it fail on my end.
> 
> Looking at the logs you sent, i can see : [Warning] RVLS not detected
> 
> This mean the simulation are running, but aren't tested against the RVLS checker. You probably need to compile ext/rvls.
> 
> See https://github.com/SpinalHDL/rvls/tree/94f850e5f7c9a36e5aceb3810d658fffae55b497#how-to-compile =>
> 

You may forgot the current spike does not support H extension, so it will be always failed. XD. (It also does not support SSTC extension, XD)
Maybe you can point the way for bumping the spike (especially the thing you change) and we can have a try.

Anyway, I think most of thing are fine since this is the riscv-arch-test has validated itself, so I guess the only problem is bumping the spike.

> ```shell
> # Compile riscv-isa-sim (spike), used as a golden model during the sim to check the dut behavior (lock-step)
> cd VexiiRiscv/ext/riscv-isa-sim
> mkdir build
> cd build
> # Optionally add CFLAGS='-g -O0' CXXFLAGS='-g -O0'
> ../configure --prefix=$RISCV --enable-commitlog  --without-boost --without-boost-asio --without-boost-regex
> make -j$(nproc)
> cd ../..
> 
> # Compile RVLS
> cd rvls
> make -j$(nproc)
> ```
> 
> Then, when you will run a simulation, as soon as there is a missmatch between the sim and the golden reference, it will stop the simulation and fail. 

> That is maybe why you had endless runtime with buildroot enable, maybe one simulation locked up, which would normaly have been detected by RVLS.

In fact it is not an endless run, but a long time run and it finally successed. This is simply because my tester has a bad single core performance and force on mult-cores.

> 
> Good luck :D

inochisa:
@Dolu1990. let me try a vibe locally first.

Dolu1990:
Note, one first step would be to try with : 
https://github.com/project-inochi/VexiiRiscv/blob/5715d2362ca9546d48020874b9316fcd1ccf5578/src/test/scala/vexiiriscv/tester/Regression.scala#L136 modified without "--with-isa g,h"
To be sure the base ISA without the heavy changes of the PR are still fine. (i ran a few test localy, it seems fine without the hypervisor option, but i don't have enough RAM to run it fast)

Then once this is fine, moving forward with hypervisor option ^^

inochisa:
> Note, one first step would be to try with : https://github.com/project-inochi/VexiiRiscv/blob/5715d2362ca9546d48020874b9316fcd1ccf5578/src/test/scala/vexiiriscv/tester/Regression.scala#L136 modified without "--with-isa g,h" To be sure the base ISA without the heavy changes of the PR are still fine.

Great. Thanks for this information.

> (i ran a few test localy, it seems fine without the hypervisor option, but i don't have enough RAM to run it fast)

I have enough RAM for each tester, but the single core performance is pretty bad. XD.

> 
> Then once this is fine, moving forward with hypervisor option ^^

I have tried a vibe coding on the RVLS/spike and got a small change for upstream. Let me review if it works prefect and the patch easy enough to understand. (At least I think some of them should be changed at least for now.....)
Then I will test both of them and send a PR for it. Thanks.

inochisa:
Hi, @Dolu1990 
Could you try the latest PR with the following rvls/spike?
https://github.com/project-inochi/rvls/tree/upstream/vexiiriscv
https://github.com/project-inochi/riscv-isa-sim/tree/upstream/vexiiriscv

The pass log is here.
[vexii.log](https://github.com/user-attachments/files/28618313/vexii.log)

inochisa:
Update, I have rebased to the latest dev.

Dolu1990:
Thanks ^^
I will try.

Note, i pushed in dev-ext some change which comment out / disable some Imsic related code, as it isn't compatible with SpinalHDL upstream : 
https://github.com/SpinalHDL/VexiiRiscv/commit/29c7752bb6c91a9392ffd383e193024126d5c013

That probably something for another PR

inochisa:
Hi @Dolu1990 , Could you try this commit:
https://github.com/SpinalHDL/VexiiRiscv/commit/9dedd82e9d909dffa1326bb809c67b987c5d2112

It is simple and I guess it is enough for masking the build error, and if not, feel free to comment them out.

Dolu1990:
Did you added modifications to ext/riscv-isa-sim or rvls, which aren't referenced in the PR ?
I get a : 
error: bad --isa option 'RV32IMF_zicntr_zihpm'. unsupported extension: zicntr
when i try the regressions

inochisa:
Hi, @Dolu1990 , could you verify the git hash of the each repo:
VexiiRiscv: `4de44230`
SpinalHDL: `6f8510cdb`
NaxSoftware: `78c7e90`
rvls: `99eba16`
riscv-isa-sim: `9451d8bd`

With the right configuration, `git status` should only show rvls and riscv-isa-sim are modified. And these are the thing I posted before.

Update, I have opened a PR SpinalHDL/rvls#10 for you to track the rvls change. If you can prepare a upstream branch for riscv-isa-sim, I am happy to open a PR for the other changes.

Dolu1990:
> rvls: 99eba16
riscv-isa-sim: 9451d8bd

Ahh that was the missmatch, thanks ^^

inochisa:
> > rvls: 99eba16
> > riscv-isa-sim: 9451d8bd
> 
> Ahh that was the missmatch, thanks ^^

FYI, the riscv-isa-sim can be found at https://github.com/project-inochi/riscv-isa-sim/tree/upstream/vexiiriscv. And the rvls change can be found in the PR.

inochisa:
Update with the latest dev.

Dolu1990:
Ahhh very weird, when compiling RVLS, i get (linux mint, gcc: 14.2.0)

In file included from /usr/include/c++/14/bits/atomic_base.h:42,
                 from /usr/include/c++/14/bits/shared_ptr_atomic.h:33,
                 from /usr/include/c++/14/memory:81,
                 from src/hart.hpp:12,
                 from src/context.hpp:5,
                 from src/context.cpp:1:
/usr/include/c++/14/bits/atomic_wait.h: In function ‘void std::__detail::__platform_wait(const _Tp*, __platform_wait_t)’:
/usr/include/c++/14/bits/atomic_wait.h:112:29: error: ‘SYS_futex’ was not declared in this scope
  112 |         auto __e = syscall (SYS_futex, static_cast<const void*>(__addr),
      |                             ^~~~~~~~~
/usr/include/c++/14/bits/atomic_wait.h: In function ‘void std::__detail::__platform_notify(const _Tp*, bool)’:
/usr/include/c++/14/bits/atomic_wait.h:125:18: error: ‘SYS_futex’ was not declared in this scope
  125 |         syscall (SYS_futex, static_cast<const void*>(__addr),

Did you had any similar issue ? On which OS / compiler are you working ?
This can be an issue on my environnement.

inochisa:
> Ahhh very weird, when compiling RVLS, i get (linux mint, gcc: 14.2.0)
> 
> In file included from /usr/include/c++/14/bits/atomic_base.h:42, from /usr/include/c++/14/bits/shared_ptr_atomic.h:33, from /usr/include/c++/14/memory:81, from src/hart.hpp:12, from src/context.hpp:5, from src/context.cpp:1: /usr/include/c++/14/bits/atomic_wait.h: In function ‘void std::__detail::__platform_wait(const _Tp*, __platform_wait_t)’: /usr/include/c++/14/bits/atomic_wait.h:112:29: error: ‘SYS_futex’ was not declared in this scope 112 | auto __e = syscall (SYS_futex, static_cast<const void*>(__addr), | ^~~~~~~~~ /usr/include/c++/14/bits/atomic_wait.h: In function ‘void std::__detail::__platform_notify(const _Tp*, bool)’: /usr/include/c++/14/bits/atomic_wait.h:125:18: error: ‘SYS_futex’ was not declared in this scope 125 | syscall (SYS_futex, static_cast<const void*>(__addr),
> 
> Did you had any similar issue ? 

In fact, I got nothing.

> On which OS / compiler are you working ? This can be an issue on my environnement.

My OS is arch linux.
With GCC 16.1.1, RISC-V GNU GCC 15.1.0 and RISC-V ELF GCC 15.2.0.

I am not sure whether it is related to the -std=c++20 in the Makefile (IIRC the spike require this). 

But your error is more like a out of sync environnement.....

Dolu1990:
> But your error is more like a out of sync environnement.....

Yes ^^ 
it is realy weird XD
I'm digging

Dolu1990:
So, got it to work on my system, had to add : 

```diff
diff --git a/src/ascii_frontend.hpp b/src/ascii_frontend.hpp
index 8d862c4..e082091 100644
--- a/src/ascii_frontend.hpp
+++ b/src/ascii_frontend.hpp
@@ -6,7 +6,7 @@
  */
 
 #pragma once
-
+#include <sys/syscall.h>
 #include <stdint.h>
 #include <string>
 #include <memory>
diff --git a/src/elf.hpp b/src/elf.hpp
index 996eab0..9276a25 100644
--- a/src/elf.hpp
+++ b/src/elf.hpp
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <sys/syscall.h>
 #include "stdio.h"
 #include "type.h"
 #include <stdexcept>
diff --git a/src/hart.hpp b/src/hart.hpp
index 82fdf23..a5243fd 100644
--- a/src/hart.hpp
+++ b/src/hart.hpp
@@ -6,7 +6,7 @@
  */
 
 #pragma once
-
+#include <sys/syscall.h>
 #include <stdint.h>
 #include <string>
 #include <memory>

```

inochisa:
Thanks, It seems like this is caused by a different header includes of different libc. I have updated the rvls.

Dolu1990:
Got : 

```
rawrr@rawrr-nb:/media/data/proj/upstream/VexiiRiscv-rvh/regression$ cd $PWD && find . -name FAIL && find . -name PASS | wc -l && find . -name FAIL | wc -l
./VexiiRiscv_1765663597_tests/buildroot/FAIL
./VexiiRiscv_563950946_tests/buildroot/FAIL
./VexiiRiscv_407100677_tests/buildroot/FAIL
./VexiiRiscv_1857588511_tests/buildroot/FAIL
./VexiiRiscv_1362383225_tests/buildroot/FAIL
./VexiiRiscv_1080580360_tests/buildroot/FAIL
./VexiiRiscv_527953930_tests/buildroot/FAIL
./VexiiRiscv_933276455_tests/buildroot/FAIL
./VexiiRiscv_1685812116_tests/buildroot/FAIL
4752
9
rawrr@rawrr-nb:/media/data/proj/upstream/VexiiRiscv-rvh/regression$ cat ./VexiiRiscv_1765663597_tests/buildroot/stdout.log 
[Progress] Start VexiiRiscv_1765663597 buildroot simulation with seed 2
[Error] Simulation failed at time=196110
INTEGER WRITE MISSMATCH DUT=2 REF=0
```

So, all failing on buildroot.
Do i need to update the buildroot images ?

inochisa:
I am not sure, but I guess it is not need. could you share the failed configuration and log? So I can check and reproduce it locally.

Dolu1990:
```
cat ./VexiiRiscv_1765663597_tests/args  && echo
--with-isa m --performance-counters 4 --lanes 1 --decoders 1 --regfile-sync --regfile-dual-ports --allow-bypass-from 1 --xlen 64 --with-btb --with-ras --without-isa m --div-ipc --div-radix 4 --with-isa m,a --with-isa m,c --with-isa b --fetch-fork-at 1 --fetch-l1-tags-read-async --debug-privileged --debug-triggers 4 --debug-triggers-lsu --debug-jtag-tap --with-lsu-l1 --lsu-l1-sets=16 --lsu-l1-ways=4 --lsu-l1-store-buffer-slots=3 --lsu-l1-store-buffer-ops=64 --lsu-l1-tags-read-async --with-fetch-l1 --lsu-l1-coherency --with-lsu-bypass --with-aligner-buffer --with-dispatcher-buffer --btb-sets 128 --btb-hash-width 6 --relaxed-btb --physical-width=36 --fetch-l1-mem-data-width-min 256 --with-isa s --with-tester-plugin

cat ./VexiiRiscv_1765663597_tests/buildroot/args  && echo
--no-stdin --ibus-ready-factor 0.5 --dbus-ready-factor 0.5 --fail-after 10000000000 --name buildroot --load-bin 0x80000000,ext/NaxSoftware/buildroot/images/rv64imac/fw_jump.bin --load-bin 0x80f80000,ext/NaxSoftware/buildroot/images/rv64imac/linux.dtb --load-bin 0x80400000,ext/NaxSoftware/buildroot/images/rv64imac/Image --load-bin 0x81000000,ext/NaxSoftware/buildroot/images/rv64imac/rootfs.cpio --fsm-getc "buildroot login:" --fsm-sleep 1000000 --fsm-putc root --fsm-putc-lr --fsm-getc # --fsm-putc "cat /proc/cpuinfo" --fsm-putc-lr --fsm-getc # --fsm-putc "echo 1+2+3*4 | bc" --fsm-putc-lr --fsm-getc # --fsm-putc micropython --fsm-putc-lr --fsm-getc ">>> " --fsm-putc "import math" --fsm-putc-lr --fsm-getc ">>> " --fsm-putc math.sin(math.pi/4) --fsm-putc-lr --fsm-getc ">>> " --fsm-putc "from sys import exit" --fsm-putc-lr --fsm-getc ">>> " --fsm-putc exit() --fsm-putc-lr --fsm-getc # --fsm-putc "ls /" --fsm-putc-lr --fsm-getc # --fsm-success
```

Note the rate of failure seems quite high. This isn't a unlucky bad config.
This seems to be CSR related. in particular INTEGER WRITE MISSMATCH DUT=2 REF=0 make me think of a mcause missmatch.

This seems to happen in opensbi (looking at the timings)

inochisa:
Thanks, I have found the problem (Just at the time you write the reply). It seems like I do not handle `mhpmcounter*` in rvls/spike correctly. So a unconsumed write leads to this fault. Let me try to fix it.

inochisa:
Hi, @Dolu1990, could you try the latest rvls change (you can find in the SpinalHDL/rvls#10 PR) and this PR.
I have fixed all the problem I found and successfully execute a buildroot test. You can check the update log in rvls PR. (Also, it is normal for a bad build in the github ci, I have changed the interface of rvls to fix problem.)

And if the test is still failed, please send the stdout.log, I have adding an snapshot feature so it is simple to check the problem (I am not sure why the dual-sim feature is broken in my machine XD, so the snapshot helped me a lot....)

Note: 
1. I suggest updating the buildroot output so we can have the same thing.
2. It is possible to test the buildroot only, I have adding some parameters to disable other tests. (but I am not sure it is possible to be used in regression mode......)

Example test command in my machine:
```bash
mill Test[].test.runMain vexiiriscv.tester.RegressionSingle \
  --with-isa=g,c,h,zihpm,zicntr --xlen=64 \
  --pmp-size=4 \
  --fetch-l1 --fetch-l1-ways=2 --fetch-l1-ways=4 --fetch-l1-mem-data-width-min=64 \
  --lsu-l1 --lsu-l1-ways=2 --with-lsu-bypass --lsu-l1-ways=4 --lsu-l1-mem-data-width-min=64 --lsu-l1-coherency \
  --relaxed-branch --allow-bypass-from=0 \
  --fetch-l1-hardware-prefetch=nl --fetch-l1-refill-count=3 \
  --lsu-l1-store-buffer-slots=4 --lsu-l1-store-buffer-ops=32 \
  --with-rvZcbm-llc \
  --without-riscv-tests \
  --without-riscv-arch-tests \
  --without-regular-tests \
  --without-jtag-tests \
  --with-free-rtos-tests-count 0 \
  --without-benchmark-tests
```

Test log:
buildroot: [stdout.log](https://github.com/user-attachments/files/29040865/stdout.log)
[regression.log](https://github.com/user-attachments/files/29040862/regression.log)

inochisa:
Update: a small update of NaxSoftware for fixing OpenSBI build in buildroot, this should be fetcted if using updated blob.
Update 2: I have double the buildroot test time, as I suffered multiple timeout

inochisa:
@Dolu1990 I have passed the test locally after a two-day running.
```
[/mnt/code/inochi/spinal/VexiiRiscv/regression]$ find . -name FAIL | wc -l
0
[/mnt/code/inochi/spinal/VexiiRiscv/regression]$ find . -name PASS | wc -l
22873
[/mnt/code/inochi/spinal/VexiiRiscv/regression]$ find . -name buildroot | wc -l
45
```

inochisa:
Update: Bump the NaxSoftware version

Dolu1990:
I'm still running the tests ^^

inochisa:
@Dolu1990 Good to know, I wonder if you can setup a upstream master branch so I can open a PR on riscv-isa-sim? I found it hard to merge it into the master branch of your repo.

Dolu1990:
Like this ? 
https://github.com/SpinalHDL/riscv-isa-sim/tree/master-upstream

inochisa:
@Dolu1990 Of course ^^
I have opened the PR  SpinalHDL/riscv-isa-sim#6 for you to track the change.
(Note, I have done a rebase to fit the master for the exception delegation change, I think it should not has some problem for this.)
Update: My local test is passed with this PR.

Dolu1990:
Finaly got to pass the tests localy aswell XD
Review WIP

Dolu1990:
Hi,

I'm not sure to understand the full purpose of the TranslatedDBusAccessPlugin ?

inochisa:
> Hi,
> 
> I'm not sure to understand the full purpose of the TranslatedDBusAccessPlugin ?

It does only one thing: hide the translation detail of GPA/SPA and provide an unified access interface. With this, the MMU plugin can access PTE without checking the current page table address is a GPA or SPA, and the TranslatedDBusAccessPlugin will take care of all the thing about this address.

inochisa:
In short: This plugin is used to mask the difference between SPA/GPA for MmuPlugin

Dolu1990:
Just tried some synthesis on efinix FPGA to see if there is regressions : 

```
VexiiRiscv upstream :
rv32im_branchPredict_cached8k8k ->
- Trion area    -> 42 Mhz LUT 3741   FF 3580
- Trion fmax   -> 103 Mhz LUT 3741   FF 3580
- Titanium area -> 146 Mhz LUT 3739   FF 3619
- Titanium fmax -> 355 Mhz LUT 3739   FF 3619
cached_branchPredict_cached8k8k_linux_mmuSync ->
- Trion   area -> 45 Mhz LUT 4403   FF 3670
- Trion  fmax  -> 74 Mhz LUT 4403   FF 3670
- Titanium area -> 135 Mhz LUT 4554   FF 3750
- Titanium -fmax> 312 Mhz LUT 4554   FF 3750

VexiiRiscv this PR :
rv32im_branchPredict_cached8k8k ->
- Trion  area  -> 45 Mhz LUT 3753   FF 3580
- Trion  fmax  -> 95 Mhz LUT 3753   FF 3580
- Titanium area -> 152 Mhz LUT 3715   FF 3619
- Titanium fmax -> 341 Mhz LUT 3715   FF 3619
cached_branchPredict_cached8k8k_linux_mmuSync ->
- Trion  area  -> 42 Mhz LUT 4456   FF 3705
- Trion  fmax  -> 73 Mhz LUT 4456   FF 3705
- Titanium area -> 141 Mhz LUT 4540   FF 3784
- Titanium  fmax -> 312 Mhz LUT 4540   FF 3784
```

This looks good on this end
Still reviewing more scala XD

Dolu1990:
Thanks :D

PR #121, opened 2026-03-03
Adapt the change of rvls to record the instruction in file backend

inochisa:
See SpinalHDL/rvls#9


Dolu1990:
The rvls submodule can be updated :)

inochisa:
Thanks, I have updated it. :)

PR #126, opened 2026-04-22
Add GETC empty check for PeripheralEmulator

inochisa:
Add a simple GETC empty check for the PeripheralEmulator, which make the support of mainline OpenSBI in the simulation easier.

Test PR: https://github.com/SpinalHDL/VexiiRiscv-RTD/pull/14
This commit is already in PR#120, I will remove it in the PR#120 if this is accepted.


Dolu1990:
Thanks ^^

PR #136, opened 2026-06-03
enable rs3 for the IntRegFile

Nik-Sch:
I am working on custom extensions and would like to add one that also has a third integer source register (similar to the `fmadd` instruction).
Is this something one could upstream as this doesn't change the generated core as far as I know and only opens the possibility to create instructions with 3 integer source registers.


Dolu1990:
Thanks ^^

PR #137, opened 2026-06-03
fix: GenerateTilelink failed without --with-supervisor

jiyuzh:
The missed `{}` of the original code forces the presence of `sei`. This can cause a `NullPointerException` during elaborations without `--with-supervisor` flag supplied.


Dolu1990:
Hooo nice XD

Thanks !

PR #138, opened 2026-06-04
Fix numerous spelling errors in comments

mrcmry:
No change in the code, so no change in the behavior.

Done by hand (using CRTL+. word propositions of spell-checker plugin of codium).

Checked with `git diff --staged` for no change behavior with mistral vibe.

Sorry for @inochisa if I cause merge issue, but I think the earliest we do this, the better for code stability.

By mistake I started this on the h-aia branch (but not sure) and I had only one file in conflict, so the merge should not be that problematic.


inochisa:
> Sorry for @inochisa if I cause merge issue, but I think the earliest we do this, the better for code stability.
> 
> By mistake I started this on the h-aia branch (but not sure) and I had only one file in conflict, so the merge should not be that problematic.

Don't worry, it is simple for me to resolve this.

Tectu:
Before merging this: You changed `"brutfocfing"` to `"brutforcing"` but I think it should be `"bruteforcing"` (or `"brute-forcing"`).

mrcmry:
> Before merging this: You changed `"brutfocfing"` to `"brutforcing"` but I think it should be `"bruteforcing"` (or `"brute-forcing"`).

Thanks, good cache. I also took the opportunity to change the to comment to approximately 80 columns, it's more readable (books or newspaper use usually less, for code it's an other story).

PR #142, opened 2026-06-19
litex: Export resolved bus interface widths

enjoy-digital:
Expose the resolved LiteX-facing pBus, mBus and coherent DMA interface widths through PythonArgsGen so the LiteX integration does not need to hard-code or guess them.

For coherent DMA, keep the default 32-bit address width but round up to 64 bits when the Vexii physical address width is above 32 bits, matching LiteX's supported DMA bus address widths and avoiding an extra 32-bit limitation on high-address configurations.

This keeps the generated Python metadata as the single interface contract consumed by LiteX.

### DIFF
--- a/src/main/scala/vexiiriscv/soc/litex/Soc.scala
+++ b/src/main/scala/vexiiriscv/soc/litex/Soc.scala
@@ -100,6 +100,9 @@ class SocConfig(){
     "imsic_s" -> 0xF1200000l,
   )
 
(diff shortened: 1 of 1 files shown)


inochisa:
I think the `axiLiteForce32` condition should also be checked. otherwise a false bus address width could be used.

Dolu1990:
thanks ^^

PR #143, opened 2026-06-20
Expose the ISA map to python arguments

inochisa:
Expose the generated ISA mapping to litex, so the litex can get the right ISA configuration


inochisa:
The litex PR enjoy-digital/litex#2481

Dolu1990:
Thanks ^^

PR #147, opened 2026-07-20
VexiiRiscvProbe: defer trap dispatch until late-completing uops are reported

### LINKED ISSUE #146: VexiiRiscvProbe dispatches traps to the co-sim backend before a late-completing uop is reported
# VexiiRiscvProbe dispatches traps to the co-sim backend before a late-completing uop is reported → false "DUT did trap on \<pc\>"

**Repo:** SpinalHDL/VexiiRiscv
**Kind:** verification-harness (probe) bug
**File:** `src/main/scala/vexiiriscv/test/VexiiRiscvProbe.scala`

## Summary

`VexiiRiscvProbe.checkTraps()` forwards a trap event to the co-sim backends(`backends.foreach(_.trap(...))`) the moment the trap FSM fires. But a uop that committed *in order* can still have an **unreported late completion** — most easily an FPU op (e.g. `fnmsub.s`) whose writeback lands a few cycles after the younger instruction that traps has already been seen by the probe. The probe then tells RVLS "trap now" *before* it tells RVLS "the FP op committed", so RVLS/Spike observes the trap one instruction early and aborts with:

```
DUT did trap on <pc>
```

This is a false positive: the hardware is correct. Only the *order* in which the probe hands events to the backend is wrong.

## How it shows up

```
0x80007310:  fnmsub.s ft0, ...     # commits in order, COMPLETED a few cycles later
0x80007314:  ecall                 # traps (cause=11) while fnmsub writeback still pending
```

VCD confirmation that the RTL is correct:
- At the trap cycle, the `CsrRam` write to `mepc` (`csrram[0xb]`) is  **`0x80007314`** — the `ecall` PC, i.e. the architecturally-correct `mepc` (not the `fnmsub` PC `0x80007310`).
- The `fnmsub` is in ctrl4 with `COMMIT=1, COMPLETED=0`: hardware *did* commit it, as a late-completing FP op.
- After the fix below (which touches **zero RTL**), the same program runs clean in lock-step: the `ft0` writeback + `fnmsub` commit are reported to RVLS *before* the trap, and Spike agrees on the FP result and the `mepc`.

## Root cause

`checkTraps()` unconditionally dispatches the trap. The retire/report stream (`checkCommits`) may still be behind the trap because a committed uop hasn't `done` its late completion yet.

## Proposed fix

Defer the trap until the retire pointer reaches the trap's `microOpAllocPtr` snapshot, so backend event order stays architectural. A never-completing uop still trips the existing liveness `simFailure`, so a *genuinely* dropped instruction keeps surfacing (this only masks the false positive, not real bugs).

```diff
@@ class VexiiRiscvProbe
     var commits = 0l
+    val pendingTraps = mutable.Queue[(Boolean, Int, Int)]()

@@ checkCommits
+      def flushDueTraps(): Unit = {
+        while (hart.pendingTraps.nonEmpty && hart.pendingTraps.head._3 == hart.microOpRetirePtr) {
+          val (interrupt, cause, _) = hart.pendingTraps.dequeue()
+          backends.foreach(_.trap(hart.hartId, interrupt, cause))
+        }
+      }
+      flushDueTraps()
       while (hart.microOpRetirePtr != hart.microOpAllocPtr && hart.microOp(hart.microOpRetirePtr).done) {
         ...
         hart.microOpRetirePtr = (hart.microOpRetirePtr + 1) & microOpIdMask
+        flushDueTraps()
       }
@@ checkTraps
   def checkTraps(): Unit = {
     for(trap <- proxies.trap) if(trap.fire.toBoolean){
-      backends.foreach(_.trap(hartsIds(trap.hartId), trap.interrupt.toBoolean, trap.cause.toInt))
+      val hart = harts(trap.hartId)
+      if (hart.microOpRetirePtr == hart.microOpAllocPtr && hart.pendingTraps.isEmpty) {
+        backends.foreach(_.trap(hartsIds(trap.hartId), trap.interrupt.toBoolean, trap.cause.toInt))
+      } else {
+        hart.pendingTraps.enqueue((trap.interrupt.toBoolean, trap.cause.toInt, hart.microOpAllocPtr))
+      }
     }
   }
```

wndmll643:
Fixes #146

`checkTraps()` dispatched a trap to the co-sim backends the moment the trap FSM fired, even when an older, in-order-committed uop (e.g. an FPU op with a late writeback) had not yet been reported through `checkCommits()`. RVLS then observed the trap one instruction early and aborted with a false `DUT did trap on <pc>`, although the RTL behavior (including `mepc`) was architecturally correct.

This PR queues such a trap together with a snapshot of `microOpAllocPtr`, and dispatches it from the retire stream once `microOpRetirePtr` reaches that snapshot, so the backends observe commits and traps in architectural order. When nothing is in flight, the trap is dispatched immediately as before. A uop that never completes still trips the existing liveness `simFailure`, so genuinely dropped instructions keep surfacing.

Touches only the probe — zero RTL change. Validated on the reproducer from #146 (`fnmsub.s` followed by `ecall`): with the fix, the FP writeback and commit are reported to RVLS before the trap and the program runs clean in lock-step.

### DIFF
--- a/src/main/scala/vexiiriscv/test/VexiiRiscvProbe.scala
+++ b/src/main/scala/vexiiriscv/test/VexiiRiscvProbe.scala
@@ -197,6 +197,7 @@ class VexiiRiscvProbe(cpu : VexiiRiscv, kb : Option[konata.Backend], var withRvl
 
     var commits = 0l
 
+    val pendingTraps = mutable.Queue[(Boolean, Int, Int)]()
 
     val jbStats = mutable.HashMap[Long, JbStats]()
     val branchStats = new JbStats()
@@ -590,6 +591,13 @@ class VexiiRiscvProbe(cpu : VexiiRiscv, kb : Option[konata.Backend], var withRvl
         simFailure(f"Vexii hasn't committed anything for too long, $status")
       }
 
+      def flushDueTraps(): Unit = {
+        while (hart.pendingTraps.nonEmpty && hart.pendingTraps.head._3 == hart.microOpRetirePtr) {
+          val (interrupt, cause, _) = hart.pendingTraps.dequeue()
+          backends.foreach(_.trap(hart.hartId, interrupt, cause))
+        }
+      }
+      flushDueTraps()
       while (hart.microOpRetirePtr != hart.microOpAllocPtr && hart.microOp(hart.microOpRetirePtr).done) {
         import hart._
         val uopId = hart.microOpRetirePtr
@@ -636,6 +644,7 @@ class VexiiRiscvProbe(cpu : VexiiRiscv, kb : Option[konata.Backend], var withRvl
           uop.clear()
         }
         hart.microOpRetirePtr = (hart.microOpRetirePtr + 1) & microOpIdMask
+        flushDueTraps()
       }
     }
   }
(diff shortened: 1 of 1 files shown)


Dolu1990:
I'm working on getting the CI back on track ^^

Dolu1990:
Thanks :D

PR #149, opened 2026-07-20
Add Smcntrpmf/Shlcofideleg extension support

inochisa:
Currently the code of performance counter plugin is restructured by the previous #120, now we can add Smcntrpmf support with only code moving. Also add Shlcofideleg extension as improve the hypervisor support.


inochisa:
@Dolu1990 I guess this CI requires a restart, it is killed by the github XD

Dolu1990:
Thanks ^^

PR #150, opened 2026-07-20
Add ASID support for the MMU

inochisa:
This PR add basic ASID and global bit support for the MMU and SFENCE.VMA/HFENCE.VVMA. It can be enabled by setting asidWidth > 0.

### DIFF
--- a/src/main/scala/vexiiriscv/Param.scala
+++ b/src/main/scala/vexiiriscv/Param.scala
@@ -854,7 +854,8 @@ class ParamSimple() {
       case true => plugins += new vexiiriscv.memory.MmuPlugin(
         spec = if (xlen == 32) MmuSpec.sv32 else MmuSpec.sv39,
         physicalWidth = physicalWidth,
-        asidWidth = asidWidth
+        asidWidth = asidWidth,
+        withGuestSfenceCheck = withHypervisor
       )
     }
     (withRvh && withMmu) match {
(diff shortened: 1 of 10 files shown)


inochisa:
Not very sure how to support VMID, as MMU and SMMU both needs to record this. maybe make the VMID field global/a separate MMU parameter plugin? I am not sure whether is better.

Dolu1990:
> Not very sure how to support VMID, as MMU and SMMU both needs to record this. maybe make the VMID field global/a separate MMU parameter plugin? I am not sure whether is better.

If that is a very simple storage CSR, can be placed in the PrivilegedPlugin, as long as it take a generation parameter to be generated.

inochisa:
> > Not very sure how to support VMID, as MMU and SMMU both needs to record this. maybe make the VMID field global/a separate MMU parameter plugin? I am not sure whether is better.
> 
> If that is a very simple storage CSR, can be placed in the PrivilegedPlugin, as long as it take a generation parameter to be generated.

This field is just part of the `hgatp` (something like `satp`), but it only contains the information for the VMID for the MMU/shadow MMU.

Although moving this field may give different for shadow MMU, but it seems like this is the only way, I think it is fine for me to move this to the PrivilegePlugin. If you think this is good enough, feel free to merge it. I will open a new PR for the VMID support. ^^

Dolu1990:
> If you think this is good enough, feel free to merge it. I will open a new PR for the VMID support. ^^

Looks good to me ^^

Issue #139, opened 2026-06-07
Is software prefetch (Zicbop from CMO) fully or partially implemented ?


mrcmry:
@Dolu1990 

It seems from the code that PREFETCH.R and .W are implemented in LSU but not there is no trace of PREFETCH.I.

Is PREFETCH.I a no-op and do not lead to invalid instruction trap ? In this case Zicbop is probably supported (not optimal because not efficient for .I but .R and .W are probably more important). Is this still experimental or do not work with certain configuration (multicore with cache coherency for example) .

If this is fully supported, should we not document this in the README.md and in Param.scala  more explicitly (I can do this).

PR #151, opened 2026-07-20
Add multiple core Zicbom support when using with cache

inochisa:
With recently SpinalHDL update, we already have the flushable cache support. So the VexiiRiscv now can enable zicbom in multiple core mode when enabling cache.

Note: Quote #139 as it improve the current implementation


Dolu1990:
Thanks :D

PR #153, opened 2026-07-31
Add unified extension manager for extension resolve.

inochisa:
It is something hard to check the extension support in VexiiRiscv currently. As this project is divided into plugin, the plugin is more likely to handle the extension check by itself, instead of require some support matrix from a global database. This means we need to duplicate some dependency check, and it is hard to use the support matrix in the early state (like the ParamSimple).

To simplify this, add an standalone extension manager to manage the extension globally, and simplify the generation of the ISA string support.


Dolu1990:
Thanks :D

PR #159, opened 2026-08-06
fpu: fix FP compare treating subnormal operand as zero (feq/fle/flt)

### LINKED ISSUE #158: FP comparisons (`feq.s`/`fle.s`/`flt.s`) treat subnormal operands as equal to zero
### Summary

VexiiRiscv's floating-point comparison unit mis-detects a **subnormal** operand as **zero**, so a
comparison between zero and a subnormal returns the wrong result. Spike (golden reference)
computes them correctly.

### Observed behavior (both confirmed with minimal repros)

| Comparison | operands | VexiiRiscv | Correct (IEEE / Spike) |
|---|---|---|---|
| `feq.s s1, f4, f6` | `+0.0`, `+subnormal (0x00400000)` | **1 (equal)** | 0 (not equal) |
| `fle.s s1, f4, f6` | `−0.0 (0x80000000)`, `−subnormal (0x80400000)` | **1 (≤)** | 0 (`0 ≤ −tiny` is false) |

A subnormal is a nonzero value; treating it as zero makes `feq`/`fle`/`fge` return "equal/≤" when
one operand is zero and the other is a subnormal.

### Root cause (likely)

`src/main/scala/vexiiriscv/execute/fpu/FpuCmpPlugin.scala` — the zero-handling path:

```scala
val bothZero = insert(RS1_FP.isZero && RS2_FP.isZero)         // ~line 103
...
rs1AbsSmaller.setWhen(RS1_FP.isZero)                          // ~line 111
rs1AbsSmaller.clearWhen(RS2_FP.isZero)
...
val CMP_RESULT = insert(rs1Smaller && !bothZero && LESS || (rs1Equal || bothZero) && EQUAL)  // ~line 123
```

If a subnormal operand is flagged `isZero`, then `bothZero` becomes true when the other operand is
zero, and `CMP_RESULT` returns the EQUAL branch (→ 1). The internal `FloatUnpacked` note
(`Interface.scala:33`) says subnormals are represented via an extended exponent rather than
flushed, so the fault appears to be the `isZero` classification (a subnormal has a nonzero
mantissa and must not be `isZero`), not a value flush.

### Minimal reproducer

```asm
    # (M-mode, mstatus.FS enabled)
    li      t0, 0x00000000
    fmv.w.x f4, t0            # f4 = +0.0
    li      t0, 0x00400000
    fmv.w.x f6, t0            # f6 = +subnormal (nonzero)
    feq.s   s1, f4, f6        # VexiiRiscv -> 1 ; correct -> 0
```

(A ready-to-run ELF with this code spliced into a bootable image is available; RVLS lock-step
flags `INTEGER WRITE MISSMATCH DUT=0x1 REF=0x0` at the `feq.s`/`fle.s`.)

### Spec basis

RISC-V F/D mandate full IEEE-754 support including subnormals; subnormals are distinct from zero,
so `feq.s(0, subnormal)` must be 0 and `fle.s(−0.0, −subnormal)` must be 0. The current results
are non-compliant.

### Suggested fix

Ensure the comparison's zero detection uses a true `isZero` (exponent minimal **and** mantissa
zero) so subnormals are not classified as zero; the ordering path already handles subnormal
magnitudes via the extended exponent.

wndmll643:
Fixes #158

## Problem

FP comparisons (`feq.s`/`fle.s`/`flt.s`, and the `.d` forms) treat a **subnormal** operand as
**zero**, so a comparison between zero and a subnormal returns the wrong result:

| Comparison | operands | before | correct (IEEE / Spike) |
|---|---|---|---|
| `feq.s s1,f4,f6` | `+0.0`, `+subnormal (0x00400000)` | **1** | 0 |
| `fle.s s1,f4,f6` | `−0.0 (0x80000000)`, `−subnormal (0x80400000)` | **1** | 0 |

## Root cause

`FpuCmpPlugin`'s `rs1Equal` compared `sign`, `exponent` and `mantissa` but **not** the recoded
`FloatMode`:

```scala
val rs1Equal = insert(RS1_FP.sign === RS2_FP.sign && expEqual && RS1_FP.mantissa === RS2_FP.mantissa)
```

Subnormals are recoded to `NORMAL` with an extended exponent. For inputs such as `0x00400000` the
recoded `exponent`+`mantissa` coincide with those of zero (`mode = ZERO`), so `rs1Equal` reported
equality across differing modes and `CMP_RESULT` took the EQUAL branch (→ 1).

## Fix

Add a mode-equality term so operands with different modes are never treated as equal (diagnosis and
one-line fix by @Dolu1990 in #158):

```scala
val rs1Equal = insert(RS1_FP.sign === RS2_FP.sign && expEqual && RS1_FP.mantissa === RS2_FP.mantissa && RS1_FP.mode === RS2_FP.mode)
```

This can only *remove* spurious equalities: two numerically-equal FP values always share the same
mode, so no legitimate comparison changes.

## How it was found / reproduced

Coverage-guided lock-step fuzzing (rv64gc programs vs Spike/RVLS). Minimal repro (M-mode,
`mstatus.FS` on, `fcsr=0`, `fpuIgnoreSubnormal=false`):

```asm
    li      t0, 0x80000000
    fmv.w.x f4, t0            # -0.0
    li      t0, 0x80400000
    fmv.w.x f6, t0            # -subnormal
    fle.s   s1, f4, f6        # was 1, correct 0
```

Maintainer confirmed the patch makes the reported test pass.

### DIFF
--- a/src/main/scala/vexiiriscv/execute/fpu/FpuCmpPlugin.scala
+++ b/src/main/scala/vexiiriscv/execute/fpu/FpuCmpPlugin.scala
@@ -102,7 +102,7 @@ class FpuCmpPlugin(val layer : LaneLayer,
 
       val bothZero = insert(RS1_FP.isZero && RS2_FP.isZero)
       val expEqual = insert(RS1_FP.exponent === RS2_FP.exponent)
-      val rs1Equal = insert(RS1_FP.sign === RS2_FP.sign && expEqual && RS1_FP.mantissa === RS2_FP.mantissa)
+      val rs1Equal = insert(RS1_FP.sign === RS2_FP.sign && expEqual && RS1_FP.mantissa === RS2_FP.mantissa && RS1_FP.mode === RS2_FP.mode)
       val rs1ExpSmaller = insert(RS1_FP.exponent < RS2_FP.exponent)
       val rs1MantissaSmaller = insert(RS1_FP.mantissa < RS2_FP.mantissa)
 


wndmll643:
Superseded by #160 (clean single-commit history). Same one-line fix.

Issue #9, opened 2024-03-14
Run a simple bare metal program?


nachiket:
Is this a better repo to use for simple experiments than the original? 

I see there's a pointer to running the ELF simulation at this URL: https://spinalhdl.github.io/VexiiRiscv-RTD/master/VexiiRiscv/HowToUse/index.html

How/where do I supply my own code to generate an ELF for functional and performance evaluation?

Dolu1990:
> Is this a better repo to use for simple experiments than the original?

original meaning VexRiscv ? It is probably a bit early if you think about FPGA / SoC, but for raw CPU simulation, i think it is good.

I would say here : 
ext/NaxSoftware/baremetal
you can add a new "project" there, then cd your project and "make rv32ima" for instance

note that by default, most option / optimization for the CPU architecture are disabled.

PR #161, opened 2026-08-08
Add AIA/IMSIC simulation support to the TestBench

inochisa:
It is hard to check AIA related function as it is complex and require lots of external controller (See litex integration). This PR add an initial TestBench support for the VexiiRiscv.

The design of this PR is divided into three parts:
1. Add a new generation option to allow generate VexiiRiscv without external interrupt line. This is useful for simulation as when enable IMSIC, the interrupt is mainly come from IMSIC instead of the external interrupt line.
2. Add a `IoDeviceManager` to simpify the management of the IO device.
3. Add a new simulated IMSIC device `ImsicPeripheralEmulator`,

Behavior mismatch:
There is a behavior mismatch between TestBench and expected RTL. The write to IMSIC in RTL should return after the interrupt is received by the IMSIC File. However, the TestBench return immediately after the check. This should be safe as the MSI is always an async thing. And the design of IMSIC supports buffer between the bus and the file to reduce the soc bus pressure.

Note: 
1. No rvls support because the upstream PR https://github.com/riscv-software-src/riscv-isa-sim/pull/2342 is still under review. I will add rvls support after it is merged. And also no baremetal test as both sail and ACT has no test for this.. 
3. For linux test, although it is better to add some regression test for the AIA/IMSIC. I found it is hard to add a devicetree in a extensive way, because it is much better to reuse the supervisor blob for the test. As for some preparation, I seek for some suggestion for the test improvement. (Maybe add a new script for DTB generation separately?)

### DIFF
--- a/src/main/scala/vexiiriscv/Param.scala
+++ b/src/main/scala/vexiiriscv/Param.scala
@@ -610,6 +610,7 @@ class ParamSimple() {
     if (withMmu) r += s"asid$asidWidth"
     if (withIterativeShift) r += "isft"
     if (extension.withDiv) r += s"d${divRadix}${divImpl}${if(divArea)"Area" else ""}"
+    if (privParam.withImsic) r += s"if${privParam.imsicInterrupts}gf${privParam.guestExternalInterruptFiles}"
     if (privParam.withDebug) r += s"pdbg"
     if (privParam.withExternalInterrupt) r += "nei"
     if (embeddedJtagTap) r += s"jtagt"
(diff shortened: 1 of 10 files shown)


Dolu1990:
Thanks ^^

> > For linux test, although it is better to add some regression test for the AIA/IMSIC. I found it is hard to add a devicetree in a extensive way, because it is much better to reuse the supervisor blob for the test. As for some preparation, I seek for some suggestion for the test improvement. (Maybe add a new script for DTB generation separately?)

Ahh this is partialy addressed via https://github.com/SpinalHDL/NaxSoftware/pull/9 ?

inochisa:
> Thanks ^^
> 
> > > For linux test, although it is better to add some regression test for the AIA/IMSIC. I found it is hard to add a devicetree in a extensive way, because it is much better to reuse the supervisor blob for the test. As for some preparation, I seek for some suggestion for the test improvement. (Maybe add a new script for DTB generation separately?)
> 
> Ahh this is partialy addressed via [SpinalHDL/NaxSoftware#9](https://github.com/SpinalHDL/NaxSoftware/pull/9) ?

No thing to address, this PR is for #169. And we should after bumping the buildroot blob version after merging #169.

PR #164, opened 2026-08-22
Fix guest time access traps

flaviens:
Guest `time` denial always raised cause 2. Raise cause 22 when `hcounteren.TM` is the blocking gate.

Verification: `sbt compile`.

### DIFF
--- a/src/main/scala/vexiiriscv/execute/CsrService.scala
+++ b/src/main/scala/vexiiriscv/execute/CsrService.scala
@@ -235,7 +235,7 @@ class CsrHartApi(csrService: CsrService, hartId : Int){
     }
   }
 
-  def virtualizeCsr(csrFilter : Any, cond: Bool) = csrService.onDecode(csrFilter) {
+  def isVirtualAccess(csrFilter : Any, cond: Bool) = csrService.onDecode(csrFilter) {
     when(csrService.accessHartId(hartId)) {
       when (cond) {
         csrService.bus.decode.doVirtual()
--- a/src/main/scala/vexiiriscv/execute/CsrService.scala
+++ b/src/main/scala/vexiiriscv/execute/CsrService.scala
@@ -235,6 +235,14 @@ class CsrHartApi(csrService: CsrService, hartId : Int){
     }
   }
 
+  def virtualizeCsr(csrFilter : Any, cond: Bool) = csrService.onDecode(csrFilter) {
+    when(csrService.accessHartId(hartId)) {
+      when (cond) {
+        csrService.bus.decode.doVirtual()
+      }
+    }
+  }
+
   def onRead(csrFilter : Any, onlyOnFire : Boolean)(body : => Unit) = csrService.onRead(csrFilter, onlyOnFire){
     when(csrService.writingHartId(hartId)){ body }
   }
--- a/src/main/scala/vexiiriscv/misc/PrivilegedPlugin.scala
+++ b/src/main/scala/vexiiriscv/misc/PrivilegedPlugin.scala
@@ -1257,25 +1257,32 @@ class PrivilegedPlugin(val p : PrivilegedParam, val hartIds : Seq[Int]) extends 
         }
 
         val guest = p.withHypervisor generate new Area {
-          val allowVirtualSupervisor = m.counteren.tm && h.counteren.tm
+          val allowHost = m.counteren.tm
+          val allowVirtualSupervisor = h.counteren.tm
           val allowVirtualUser = privilege === PrivilegeMode.VS || s.counteren.tm
           val accessable = allowVirtualSupervisor && allowVirtualUser
-          val virtual = m.counteren.tm && !h.counteren.tm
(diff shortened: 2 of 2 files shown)


src/main/scala/vexiiriscv/execute/CsrService.scala
```diff
@@ -235,6 +235,14 @@ class CsrHartApi(csrService: CsrService, hartId : Int){
     }
   }
 
+  def virtualizeCsr(csrFilter : Any, cond: Bool) = csrService.onDecode(csrFilter) {
```
inochisa: A better name? The thing you do is make the access as a virtual access, not the CSR itself is a virtual CSR.
maybe `isVirtualAccess`?

src/main/scala/vexiiriscv/misc/PrivilegedPlugin.scala
```diff
@@ -1268,10 +1269,13 @@ class PrivilegedPlugin(val p : PrivilegedParam, val hartIds : Seq[Int]) extends
               api.read(rdtime(63 downto 32), GuestCsrFilter(CSR.UTIMEH))
               api.allowCsr(GuestCsrFilter(CSR.UTIME), accessable)
               api.allowCsr(GuestCsrFilter(CSR.UTIMEH), accessable)
+              api.virtualizeCsr(GuestCsrFilter(CSR.UTIME), virtual)
+              api.virtualizeCsr(GuestCsrFilter(CSR.UTIMEH), virtual)
```
inochisa: Please make the `GuestCsrFilter(CSR.UTIME/UTIMEH)` a variable, so the builder can know it is the same CSR filter. `GuestCsrFilter(CSR.UTIME) == GuestCsrFilter(CSR.UTIME)` is false actually, so it is better to use a variable to let the build know it is the same filter for some optimization.

Also, I think you should do a host denied check if `m.counteren.tm` is not set.

src/main/scala/vexiiriscv/misc/PrivilegedPlugin.scala
```diff
@@ -1268,10 +1269,13 @@ class PrivilegedPlugin(val p : PrivilegedParam, val hartIds : Seq[Int]) extends
...
             }
             case 64 => {
               api.read(rdtime, GuestCsrFilter(CSR.UTIME))
               api.allowCsr(GuestCsrFilter(CSR.UTIME), accessable)
+              api.virtualizeCsr(GuestCsrFilter(CSR.UTIME), virtual)
```
inochisa: The same as 32-bit CSR

inochisa (changes requested):

flaviens:
Thanks, I’ll address these points and update both RV32 and RV64.

src/main/scala/vexiiriscv/misc/PrivilegedPlugin.scala
```diff
@@ -1257,25 +1257,32 @@ class PrivilegedPlugin(val p : PrivilegedParam, val hartIds : Seq[Int]) extends
...
+              val timeFilter = GuestCsrFilter(CSR.UTIME)
+              val timehFilter = GuestCsrFilter(CSR.UTIMEH)
+              api.read(rdtime(31 downto 0), timeFilter)
+              api.read(rdtime(63 downto 32), timehFilter)
+              api.allowHostCsr(timeFilter, allowHost)
+              api.allowHostCsr(timehFilter, allowHost)
+              api.allowCsr(timeFilter, accessable)
+              api.allowCsr(timehFilter, accessable)
+              api.isVirtualAccess(timeFilter, virtual)
+              api.isVirtualAccess(timehFilter, virtual)
```
inochisa, line 1278: I think this can be removed, it should hit the first condition If my memory is right XD.
https://github.com/SpinalHDL/VexiiRiscv/blob/0340addf98e1d3f0d41f97e66933a2713f884b54/src/main/scala/vexiiriscv/misc/PrivilegedPlugin.scala#L1324-L1335

src/main/scala/vexiiriscv/misc/PrivilegedPlugin.scala
```diff
@@ -1257,25 +1257,32 @@ class PrivilegedPlugin(val p : PrivilegedParam, val hartIds : Seq[Int]) extends
...
             }
             case 64 => {
-              api.read(rdtime, GuestCsrFilter(CSR.UTIME))
-              api.allowCsr(GuestCsrFilter(CSR.UTIME), accessable)
-              api.virtualizeCsr(GuestCsrFilter(CSR.UTIME), virtual)
+              val timeFilter = GuestCsrFilter(CSR.UTIME)
+              api.read(rdtime, timeFilter)
+              api.allowHostCsr(timeFilter, allowHost)
+              api.allowCsr(timeFilter, accessable)
+              api.isVirtualAccess(timeFilter, virtual)
```
inochisa, line 1285: See the comment above

inochisa (changes requested):

PR #166, opened 2026-08-22
Clear mtval2 on access faults

flaviens:
`mtval2` received a guest address on access faults. Write it only for guest-page faults.

Verification: `sbt compile`.

### DIFF
--- a/src/main/scala/vexiiriscv/misc/TrapPlugin.scala
+++ b/src/main/scala/vexiiriscv/misc/TrapPlugin.scala
@@ -826,7 +826,7 @@ class TrapPlugin(val trapAt : Int, val recordHtinst : Boolean) extends FiberPlug
               goto(JUMP) // improvement: shave one cycle
               when(satsPorts.refill.rsp.pageFault || satsPorts.refill.rsp.accessFault) {
                 pending.state.exception := True
-                when(!satsPorts.refill.rsp.accessFault) {
+                when(satsPorts.refill.rsp.pageFault) {
                   buffer.trap.tval2 := satsPorts.refill.rsp.address.dropLow(2).asBits.resized
                 }
                 switch(satsPorts.refill.rsp.accessFault ## pending.state.arg(1 downto 0)){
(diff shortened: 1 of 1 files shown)


src/main/scala/vexiiriscv/misc/TrapPlugin.scala
```diff
@@ -826,7 +826,9 @@ class TrapPlugin(val trapAt : Int, val recordHtinst : Boolean) extends FiberPlug
               goto(JUMP) // improvement: shave one cycle
               when(satsPorts.refill.rsp.pageFault || satsPorts.refill.rsp.accessFault) {
                 pending.state.exception := True
-                buffer.trap.tval2 := satsPorts.refill.rsp.address.dropLow(2).asBits.resized
+                when(!satsPorts.refill.rsp.accessFault) {
+                  buffer.trap.tval2 := satsPorts.refill.rsp.address.dropLow(2).asBits.resized
+                }
```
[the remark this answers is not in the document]
flaviens: Fixed.

src/main/scala/vexiiriscv/misc/TrapPlugin.scala
```diff
@@ -774,8 +774,8 @@ class TrapPlugin(val trapAt : Int, val recordHtinst : Boolean) extends FiberPlug
                 atsPorts.refill.rsp.ready := True
                 pending.state.exception := True
                 if (priv.p.withHypervisor) when(atsPorts.isGuestRefill) {
-                  buffer.trap.tval2 := atsPorts.refill.rsp.address.dropLow(2).asBits.resized
                   when (atsPorts.refill.rsp.guestFault) {
+                    buffer.trap.tval2 := atsPorts.refill.rsp.address.dropLow(2).asBits.resized
```
inochisa, line 778: MmuPlugin always reports right fault address, so why for this fix? For non-guest exception, it is always 0, so dropping 2 bits does not hurts

src/main/scala/vexiiriscv/misc/TrapPlugin.scala
```diff
@@ -826,7 +826,9 @@ class TrapPlugin(val trapAt : Int, val recordHtinst : Boolean) extends FiberPlug
               goto(JUMP) // improvement: shave one cycle
               when(satsPorts.refill.rsp.pageFault || satsPorts.refill.rsp.accessFault) {
                 pending.state.exception := True
-                buffer.trap.tval2 := satsPorts.refill.rsp.address.dropLow(2).asBits.resized
+                when(satsPorts.refill.rsp.pageFault) {
+                  buffer.trap.tval2 := satsPorts.refill.rsp.address.dropLow(2).asBits.resized
+                }
```
inochisa, line 831: The same as above

flaviens:
The address is correct but `mtval2` must remain zero for non-guest-page faults (the guard is here for doing that distinction)

PR #168, opened 2026-08-22
Check first-stage permissions during TLB refill

flaviens:
First-stage PTE permission checks were only applied on TLB hits. A cold two-stage access could therefore refill the VS-stage translation and consult the G stage before noticing a VS-stage permission fault, making the exception cause depend on TLB residency.

Carry the access privilege context into the refill request and apply the same U/R/W/X, SUM, MXR, MPRV, plus SPVP checks to the leaf response. A first-stage permission fault now wins before a G-stage refill is attempted.

Verification: sbt compile; RV64 elaboration with and without the H extension; full-core Verilator probe. A simultaneous VS write-permission fault plus invalid G-stage mapping now reports store page fault (0x0f) for cold/warm/cold/warm runs. The control with a valid VS mapping plus invalid G-stage mapping remains store guest-page fault (0x17).

### DIFF
--- a/src/main/scala/vexiiriscv/memory/MmuPlugin.scala
+++ b/src/main/scala/vexiiriscv/memory/MmuPlugin.scala
@@ -468,6 +468,9 @@ class MmuPlugin(var spec : MmuSpec,
       val storageOhReg = Reg(Bits(storages.size bits))
       val storageEnable = Reg(Bool())
       val isTwoStage = Reg(Bool())
+      val forceGuest = Reg(Bool())
+      val applyMprv = Reg(Bool())
+      val permission = Reg(cloneOf(arbiter.io.output.permission))
       val isGlobal = Reg(Bool())
       val asid = Reg(Bits(asidWidth bits))
 
@@ -484,6 +487,9 @@ class MmuPlugin(var spec : MmuSpec,
           virtual := arbiter.io.output.address
           load.address := (ppn @@ spec.levels.last.vpn(arbiter.io.output.address) @@ U(0, log2Up(spec.entryBytes) bits)).resized
           isTwoStage := arbiter.io.output.indirect
+          forceGuest := arbiter.io.output.forceGuest
+          applyMprv := arbiter.io.output.applyMprv
+          permission := arbiter.io.output.permission
           isGlobal := False
           asid := priv.implementHypervisor.mux(
             Mux(arbiter.io.output.indirect, vsatp.asid, satp.asid),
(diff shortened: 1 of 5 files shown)


inochisa:
This is a wrong fix. Any first-stage check is always check in the LSU. I see no reason for this fix. If you can not give a example for this. NAK for me.

flaviens:
Hi @inochisa , thanks for looking into it! So my understanding is that ok the LSU checks it but only on a hit. If you have a miss then it goes straight from the VS refill to the G-stage refill. With VS W=0 plus an invalid G mapping I get cause 23 cold but 15 warm. Am I missing something obvious?

inochisa:
> Hi @inochisa , thanks for looking into it! So my understanding is that ok the LSU checks it but only on a hit. If you have a miss then it goes straight from the VS refill to the G-stage refill. 

I think you understand is wrong, the right thing is

LSU TLB check -> MMU refill for a miss -> redo the pipeline if no error -> LSU TLB check.

TLB is always checked in VexiiRiscv.

> With VS W=0 plus an invalid G mapping I get cause 23 cold but 15 warm. Am I missing something obvious?

Do you have an minimum baremetal for this? I think this should check the LSU instead of the MMU as 15 is not right.

Also, I guess this may be caused by you other fix. I suggest closing these PRs and merge them into one PR for a better check.

flaviens:
Hi @inochisa thanks for your patient response!
[I created a zip here](https://github.com/user-attachments/files/31415647/vexii_twostage_minimal.zip). It has a bare-metal reproducer and I also vibe-coded a dockerfile and a readme with it to ensure we are trying the same things. Can you please check whether my understanding is incorrect? I would expect to get cause 15 in both cases and it should not depend on TLB residency (am I missing a subtlety maybe? -- possible)

inochisa:
@flaviens Yes, I got the thing, the reason is the optimization refill chain in TrapPlugin

MMU_REFILL -> ATS_RSP -> SATS_RSP.

As the ATS_RSP go to SATS_RSP with no extra check, so the problem occurs.

I think a simplest fix is just always let the ATS_RSP jump JUMP state without refill SMMU, and then let the LSU send request for SMMU, however, there will be a great performance overhead for this. So I vote for the permission check in MMU, 

However, I think you should refactor the code so we only need to maintain one permission check block. Add a duplicate logic is not a good choice. @Dolu1990, any suggestions?

flaviens:
Hi @inochisa thanks for the suggestion! I'm indeed not really familiar with the design so any indications are welcome. Feel free to edit if it's simpler also, otherwise happy to help.

inochisa:
In fact I have no edit permission, but I will try to fix this locally, and if you are happy, I can send another PR to fix this.

src/main/scala/vexiiriscv/memory/MmuPlugin.scala
```diff
@@ -572,6 +578,35 @@ class MmuPlugin(var spec : MmuSpec,
...
+                             (!load.flags.U && nominalUser)
+        val readFault = permission.read && !allowRead
+        val writeFault = permission.write && !allowWrite
+        val executeFault = permission.execute && !allowExecute
+        val fault = privilegeFault || readFault || writeFault || executeFault
+      }
+
       val fetch = for((level, levelId) <- spec.levels.zipWithIndex) yield new Area{
         val pteFault = (load.exception || load.levelException(levelId)) || (levelId == 0).mux(!load.leaf, False)
         val pteReadError = load.rsp.error(0)
```
inochisa, line 612: Request for code reuse refactor

inochisa (changes requested):

flaviens:
Closing this in favor of #170, which shares the permission checker and passes the cold/warm reproducer with cause 15 in both cases.

PR #169, opened 2026-08-25
Add initial multiple core TestBench support

inochisa:
Require SpinalHDL/NaxSoftware#9 for new buildroot blob, SpinalHDL/rvls#17 for rvls support.

This PR introduces a initial multiple core TestBench for VexiiRiscv with native bus (no AXI/Wishbone/JTAG support). With this, some simple coherency test can be processed. For supporting multiple core simulation, this PR refactor the TestBench code and split them as hart-local logic/hart-shared logic. And the fix any problem occurs when trying two-core simulation test.

This PR also introduces a new registers file setter for the simulation of OpenSBI. This will simplify the support of different devicetree.


Dolu1990:
Very nice ^^
Thanks

PR #170, opened 2026-08-25
Fix MMU page fault check priority in TLB refill process

inochisa:
The improved version of #168 

This PR solves permission check error in G-stage refill process, which is a check miss in original refill process in ATS_RSP to SATS_RSP. As no permission check is performed, the PTW will think the VS-stage PTE is fine and trigger a bad check on G-stage PTE, which will not be performed if the VS-stage PTE is failed.

For vivado test on XCKU5P for gch build. only 4FF/4LUT is added, so I think adding a dedicated check is acceptable.


inochisa:
@flaviens You may want a check on this fix. It also solved the problem

flaviens:
Thanks, I’ll run the reproducer against #170 and compare the result with #168.

PR #171, opened 2026-08-27
Fix hypervisor translation correctness

flaviens:
Consolidates #164, #165, #166 and #167.

### DIFF
--- a/src/main/scala/vexiiriscv/Param.scala
+++ b/src/main/scala/vexiiriscv/Param.scala
@@ -810,7 +810,9 @@ class ParamSimple() {
     withMmu match {
       case false => plugins += new vexiiriscv.memory.StaticTranslationPlugin(physicalWidth)
       case true => plugins += new vexiiriscv.memory.MmuPlugin(
-        spec = if (xlen == 32) MmuSpec.sv32 else MmuSpec.sv39,
+        spec = if (xlen == 32) {
+          if (extension.withHypervisor) MmuSpec.sv32Physical34 else MmuSpec.sv32
+        } else MmuSpec.sv39,
         physicalWidth = physicalWidth,
         asidWidth = asidWidth,
         withGuestSfenceCheck = extension.withHypervisor
@@ -819,7 +821,7 @@ class ParamSimple() {
     (extension.withRvh && withMmu) match {
       case false => plugins += new vexiiriscv.memory.StaticTranslationPlugin(physicalWidth, 1)
       case true => plugins += new vexiiriscv.memory.ShadowMmuPlugin(
-        spec = if (xlen == 32) MmuSpec.sv32 else MmuSpec.sv39,
+        spec = if (xlen == 32) MmuSpec.sv32x4 else MmuSpec.sv39x4,
         physicalWidth = physicalWidth,
         vmidWidth = 0 /* TODO */
       )
--- a/src/main/scala/vexiiriscv/execute/CsrService.scala
+++ b/src/main/scala/vexiiriscv/execute/CsrService.scala
@@ -235,6 +235,14 @@ class CsrHartApi(csrService: CsrService, hartId : Int){
     }
   }
 
+  def isVirtualAccess(csrFilter : Any, cond: Bool) = csrService.onDecode(csrFilter) {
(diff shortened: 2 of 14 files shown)


inochisa:
Suggest checking my left comment in #164 and #167. And I think some commit can be mergered for a clear view.

Also, please drop the last two commits as you should put them in PR description.

flaviens:
Sure, on it

inochisa:
I do not think merge them into only one make the things better to understand.

If I understand correctly, You just generated the code from some AI agent, did you? If so, I have some serious questions for you: do you understand the change you made? do you review all the generated code and judge whether is is meaningless/have a good behavior? At least I see the test `GuestPhysicalWidthTest` is completely meaningless.

flaviens:
It is correct that I do not have a good understanding of the codebase. AI is heavily involved. If you want an expert on it, I'm more than happy to retract and let them do.

inochisa:
> It is correct that I do not have a good understanding of the codebase. AI is heavily involved. 

It is important that one should take responsibility for his contributions, it is similar to a paper. And if you have a hard time on writing a fix/understanding this from AI. I think leaving an issue with fully reproducible is also a good contribution, and other will remember you as a good reporter ^^.

To be honest, I also used codex a lot for developing (also in this project). But the most time I need to improve the thing a lot as even the highest model can produce many mistakes.... This is why I sent this question to you.

> If you want an expert on it, I'm more than happy to retract and let them do.

At least I think you change on the guest `TIME` csr is a good point. Others I think they need further discussion.
Anyway, it is your decision to retract or continue, but I am happy to see if you can insist and finally do a great work ^^

flaviens:
Hi @inochisa very happy to hear something positive from you! 😊  (first time I think?) Unfortunately really I cannot put the required effort here to have something I'd be happy about (indeed most of the time LLMs are good first shot for small fixes but this case is rabbitholing a bit and needs more guidance). Sorry for any inconvenience caused.

inochisa:
> Hi @inochisa very happy to hear something positive from you! 😊 (first time I think?) 

Yeap, I apology for this strict attitude. I am kind of disappointed for **non-reviewed** content (I see too many similar  low-quality things from my colleague and it is very hard for me to control myself XD.)

> Unfortunately really I cannot put the required effort here to have something I'd be happy about (indeed most of the time LLMs are good first shot for small fixes but this case is rabbitholing a bit and needs more guidance). Sorry for any inconvenience caused.

Yes, that's true. A simple fix is a good point. For guest timer CSR fix. I suggest you using the following method for a start:
1. Feed the LLM to the related spec, (for this fix, it should be the hypervisor CSR, I heavily suggest you let the LLM check the issue in upstream ISA repo, which make add some extra information.)
2. Let the LLM check all related logic for this fixes, and output a plan for how to fix it. (This step may require multiple run, as LLM usually makes a mistake).
3. Let the LLM do the fix and output a detailed explanation for the fix (this is important, as the submitter should understand his code)
4. Use an another LLM session, also feed with the related spec, and then do a review on this fixes.
5. loop 1~4 until you are satisfied
6. Some code format improvement and other minor improvement

The most important thing is that **you should need to know what are you doing**. This will make your PR convincing, and give you the courage to argue with others ^^ (Reviewer can also make mistakes, for me, I made a lot of mistakes.)

Hope you can have a good start.

flaviens:
I've been reporting the bugs, this I know how to do and happy to have argued with you 😗  the fix in a design that is unfamiliar to me is outside my territory ^^'

mrcmry:
To jump into the discussion, I think we should be careful with LLM, because they are very good a generating things that look good and plausible. It's a bit of a trap, because in open source the baseline is to suppose people are acting in good faith (and contributor using LLM are) but LLM do not have faith, they just aim at being plausible.

And being plausible is all good and needed in movies and novels, but with real hardware stuffs, you must be right, not plausible. So it's important to label correctly any information in PR and issue. Like we used (and still do) to indicate "I'm not an expert in this field", "it works on my machine but not sure it's very portable", I think we should label any text or code generated by LLM as such, and detail how we checked it is correct.

Lately someone pointed me to the following document about the subject and I find it worth the read:

https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md

PR #172, opened 2026-08-27
Fix pattern match in DispatchPlugin

ISHKAFEL09:
The pattern match here matches every RfResource irrespective of RfAccess agreement, and the filter invariably returns true. I believe this is not the intended behavior.

### DIFF
--- a/src/main/scala/vexiiriscv/schedule/DispatchPlugin.scala
+++ b/src/main/scala/vexiiriscv/schedule/DispatchPlugin.scala
@@ -175,7 +175,7 @@ class DispatchPlugin(var dispatchAt : Int,
       // Identify which RS are used by the pipeline
       val resources = ll.uops.keySet.flatMap(_.resources).distinctLinked
       val readAccess = rfaReads.filter(e => resources.exists {
-        case RfResource(_, e) => true
+        case RfResource(_, `e`._1) => true
         case _ => false
       }).values
 


src/main/scala/vexiiriscv/schedule/DispatchPlugin.scala
```diff
@@ -175,7 +175,7 @@ class DispatchPlugin(var dispatchAt : Int,
       // Identify which RS are used by the pipeline
       val resources = ll.uops.keySet.flatMap(_.resources).distinctLinked
       val readAccess = rfaReads.filter(e => resources.exists {
-        case RfResource(_, e) => true
+        case RfResource(_, `e`._1) => true
```
inochisa, line 178: I think it should be `e._1`?

ISHKAFEL09: I think the variable e in the line 178 is not interpreted as a reference to that in the line 177. Rather, it shadows that definition. Hence, this clause is actually a default, match-all clause. So this pair of backticks is required, otherwise, it won’t even compile.

inochisa: You are right, I make a mistake on scala syntax. You change is right

inochisa (approved):

PR #173, opened 2026-08-27
Allow overriding the SpinalHDL selection via system properties

dnltz:
SPINALHDL_FROM_SOURCE and SPINALHDL_PATH could only be set through the environment, and the SpinalHDL version was not configurable at all: it was hardcoded to "dev", which is not published to Maven Central. Turning SPINALHDL_FROM_SOURCE off therefore left the build asking for an artifact that cannot be resolved:

  com.github.spinalhdl:spinalhdl-core_2.12:dev  ->  404

Read all three from a JVM system property first, falling back to the existing environment variable and then to the previous default. This lets a downstream project pin the SpinalHDL it wants from a committed .sbtopts, without exporting anything into the environment. Defaults and environment variables behave exactly as before.

Also restrict the -Xplugin option to source builds. With a binary SpinalHDL the idsl plugin comes from the compilerPlugin dependency, while the option pointed into ext/SpinalHDL for a jar that nothing builds in that mode.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,10 @@
-val spinalVersion = "dev"
-val spinalHdlFromSource = sys.env.getOrElse("SPINALHDL_FROM_SOURCE", "1") == "1"
-val spinalHdlPathEnabled = sys.env.contains("SPINALHDL_PATH")
-val spinalHdlPath = sys.env.getOrElse("SPINALHDL_PATH", "./ext/SpinalHDL")
+def spinalHdlOption(prop: String, env: String, default: String) =
+  sys.props.getOrElse(prop, sys.env.getOrElse(env, default))
+
+val spinalVersion = spinalHdlOption("spinalhdl.version", "SPINALHDL_VERSION", "dev")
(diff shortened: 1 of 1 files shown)


inochisa:
I think you have misunderstand what the `SPINALHDL_FROM_SOURCE` does. As VexiiRiscv always uses upstream SpinalHDL for bug fixes and new features. It can not bind to a specific version, otherwise it can not adapt the change of SpinalHDL. This is why we need dev code. `SPINALHDL_FROM_SOURCE` provides a way to reuse local published scala artifact if you build the latest SpinalHDL locally. Also, using a version other than dev will not compile as there are some features not in the latest version. It is kind of misunderstanding, but it is worked. 

Maybe a better fix can remove linking local published artifact method in sbt build completely, like the mill build.

dnltz:
@inochisa ah, I indeed I did misunderstand that variable. I just recently switched from the VexRiscv to the Vexii and this project is over-writing the SpinalHDL version defined in my build.sbt file. I also pinned VexiiRiscv and will use whatever SpinalHDL version this project defines in the future.

Thanks for explaining to me!

inochisa:
@dnltz Thanks, and thank you for your will to submit a contribution.

PR #174, opened 2026-08-28
Optimize Timing of ZbbCountPlugin

vogma:
Hi there,

i am currently implementing the draft specification of the Packed SIMD (RVP) extension into the VexiiRiscv. When using Vexii with XLEN=64 i noticed that the current implementation of the ZbbCountPlugin in Bitmanip.scala has a very long combinatorial path. 
My Project is built for the xc7a100tcsg324-1 FPGA and my Clockrate is 100MHz. When building Vexii with rv64i_zbb i get the following results:

| Metric  | Value |
| ------------- | ------------- |
| LUT  | 3989  |
| WNS  | -0.610 ns  |
| FF  | 2557  |

I implemented the ZbbCountPlugin to use the CountLeadingZeroes/CountTrailingZeroes implementations already present. 
Results for rv64i_zbb:

| Metric  | Value |
| ------------- | ------------- |
| LUT  | 4043  |
| WNS  | 0.390  |
| FF  | 2550  |

So, in the 64-bit configuration, the implementation improves slack by a full nanosecond while only requiring 54 additional LUT's. For a minimal configuration, rv32i_zbb, the results are: 

| config  | LUT | FF | WNS |
| ------------- | ------------- | ------------- | ------------- |
| zbb32_min_upstream  | 1988  | 1733 | 0.588 |
| zbb32_min_modified  | 2006  | 1733 | 0.899 |

For rv32, the ZbbCountPlugin is not in the critical path so the timing improvements should be only placement noise. 

I think for this timing improvement the costs of 18 LUT's for rv32 and 54 LUT's in rv64 are worthwhile.

I would be happy to get opinions and feedback on this.

Cheers
Marco

### DIFF
--- a/src/main/scala/vexiiriscv/execute/Bitmanip.scala
+++ b/src/main/scala/vexiiriscv/execute/Bitmanip.scala
@@ -140,10 +140,8 @@ class ZbbLogicPlugin(val layer: LaneLayer,
 
 object ZbbCountPlugin extends AreaObject {
   val FLIP = Payload(Bool())
-  val INVERT = Payload(Bool())
   val MASK = Payload(Bool())
   val WORD = Payload(Bool())
-  val OR = Payload(Bool())
 }
 
 class ZbbCountPlugin(val layer: LaneLayer,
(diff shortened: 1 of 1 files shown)


inochisa:
Good work ^^.

After checking you change, I think this relaxed the timing of decode stage instead of the ZbbCountPlugin itself? Does vivado also report an improvement on this plugin? IIRC the decode stage is very timing tight, so you change could lead a better timing. I ask for this as an reference for me to know. Anyway, it is a good work. Thanks.

src/main/scala/vexiiriscv/execute/Bitmanip.scala
```diff
@@ -140,66 +140,60 @@ class ZbbLogicPlugin(val layer: LaneLayer,
...
-      add(RvZbx.CPOPW).srcs(SRC1.RF).decode(MASK -> False, INVERT -> False, WORD -> True, OR -> False)
+      add(RvZbx.CLZW).srcs(SRC1.RF).decode(MASK -> True, FLIP -> False, WORD -> True)
+      add(RvZbx.CTZW).srcs(SRC1.RF).decode(MASK -> True, FLIP -> True, WORD -> True)
+      add(RvZbx.CPOPW).srcs(SRC1.RF).decode(MASK -> False, FLIP -> False, WORD -> True)
     }
 
     uopRetainer.release()
 
     val count = new el.Execute(executeAt) {
-      // TODO explicitly build tree instead of the long combinatorial path...
```
inochisa, line 174: I guess you may need to keep this?

vogma:
> Good work ^^.
> 
> After checking you change, I think this relaxed the timing of decode stage instead of the ZbbCountPlugin itself? Does vivado also report an improvement on this plugin? IIRC the decode stage is very timing tight, so you change could lead a better timing. I ask for this as an reference for me to know. Anyway, it is a good work. Thanks.

Thank you.

I checked the timing report in vivado. Four paths fail and all are between the register 
`system_cpu_logic_core/execute_ctrl2_up_integer_RS1_lane0_reg[25]/C` and 
`system_cpu_logic_core/execute_ctrl3_up_lane0_integer_WriteBackPlugin_logic_DATA_lane0_reg[3-6]/D`.
All paths between them go through the ZbbCountPlugin. The worst path leaving any `decode_ctrl_*` flop is +0.637ns, so not timing critical.

src/main/scala/vexiiriscv/execute/Bitmanip.scala
vogma: But this is what [CountLeadingZeros](https://github.com/SpinalHDL/SpinalHDL/blob/3ade132e2dd90bff412885e2c2054bd98b9e4d68/lib/src/main/scala/spinal/lib/Utils.scala#L1419) already does isn't it? In my understanding this is a recursive divide-and-conquer approach which avoids the long comb path.
Also [CountOne](https://github.com/SpinalHDL/SpinalHDL/blob/3ade132e2dd90bff412885e2c2054bd98b9e4d68/lib/src/main/scala/spinal/lib/Utils.scala#L384) chunks the input and combines with `reduceBalancedTree(_+_)`

inochisa: Oh, you are right. I forgot this.

inochisa:
Although I see more resources increased in my build, I think this is acceptable.

inochisa:
Thanks ^^

PR #176, opened 2026-08-30
Merge whenDouble check to improve FPU timing

inochisa:
It seems like that even the vivado can not merge the condition path, so do this by manual to get better WNS


Dolu1990:
Thanks ^^

PR #181, opened 2026-09-18
Allow to set host for TestBenchServer

inochisa:
Add option to set binded ip/socket for TestBenchServer.

As the `ServerSocketChannel` is supported after java 16. So raise the CI version for compilation.


Dolu1990:
Thanks :D

PR #182, opened 2026-09-18
Fix PMP granularity check

### LINKED ISSUE #180: pmpcfg.A=NA4 is not WARL-legalized when NAPOT is enabled (reads back an unselectable mode at granularity>4)
## Summary
With a PMP granularity greater than 4 bytes (G ≥ 1), the `NA4` address-matching
mode is not selectable, and `pmpcfg.A` (a WARL field) must read back a legal
value. VexiiRiscv, when NAPOT is enabled, **keeps and reads back `A=NA4`** after a
write of `A=NA4`, instead of legalizing it. A subsequent `pmpcfg` read therefore
returns an unselectable (non-compliant) `A` value.

## Configuration
`--pmp-size 8` with the default `granularity = 4096` (G = 10), `withNapot = true`,
`withTor = true` (RV64).

## Minimal reproducer (3 instructions, loaded at reset vector 0x80000000)
```asm
_start:
    li    t0, 0x17        # 80000000: 42dd            pmp0cfg byte: A=NA4(2), R=W=X=1
    csrw  pmpcfg0, t0     # 80000002: 3a029073        write NA4 (unselectable at G>=1 -> must WARL-legalize)
    csrr  t1, pmpcfg0     # 80000006: 3a002373        read back pmpcfg0
```
Assemble (self-contained; no attachment needed):
`riscv64-unknown-elf-gcc -march=rv64imac_zicsr -mabi=lp64 -nostdlib -nostartfiles
-Ttext=0x80000000 na4.S -o na4.elf`
Observed: `t1 = 0x17`, i.e. `pmpcfg0.A` reads back as `NA4` — the unselectable
mode is retained.

Expected: `pmpcfg.A` must not read back `NA4` when G ≥ 1; it should read a
selectable value, e.g. `0x1f` (`A=NAPOT`, with R/W/X preserved). Which selectable
value `NA4` legalizes to is implementation-defined; retaining `NA4` is not
permitted.

## Root cause
`src/main/scala/vexiiriscv/memory/PmpPlugin.scala` (~L131-134):
```scala
if(!p.withNapot){
  if(granularity > 4) when(cfgNext.kind === 2) {cfg.kind := 0}   // NA4 legalization
  when(cfgNext.kind === 3) {cfg.kind := 0}                        // NAPOT legalization
}
```
The `NA4 -> OFF` legalization (the `granularity > 4` line) is nested inside
`if(!p.withNapot)`. When NAPOT is enabled that block is not generated, so a written
`A=NA4` is never legalized. NA4 being unsupported at `granularity > 4` is
independent of whether NAPOT is supported.

Additionally, the matcher computes `isNapot = RegNext(cfg.kind(1))`; since `NA4`
(`kind=2`) has bit 1 set, a stored NA4 is matched as NAPOT at granularity 4096,
i.e. the region size differs from both NA4 and a legalized mode — so the effect is
not limited to the CSR read-back.

## Suggested fix
```scala
if(granularity > 4) when(cfgNext.kind === 2) {cfg.kind := 0}   // always: NA4 unselectable at G>=1
if(!p.withNapot) when(cfgNext.kind === 3) {cfg.kind := 0}       // NAPOT only when unsupported
```

## Spec references
- RISC-V Privileged: all PMP CSR fields are WARL; `pmpcfg` A field encoding
  0=OFF, 1=TOR, 2=NA4, 3=NAPOT.
- "When G ≥ 1, the NA4 mode is not selectable." A WARL field must read back a
  legal (here, selectable) value.

## How this differs from #145
#145 is about the cacheless LSU's page-table-walk / access **not being PMP-checked**
(a missing enforcement check). This is a separate defect in the **`pmpcfg.A` WARL
write-legalization** logic — the enforcement path is not involved; the field simply
retains an unselectable mode.

inochisa:
close #180


Dolu1990:
Thanks ^^

